PR #204, opened 2013-12-03
Feature/layer based tilestore

hwbllmnn:
This adds a tile store that is based on any ordinary deegree layer. Together with the file system tile store enhancement this can be used to create a tile cache for any configured layer on demand.


tfr42:
To reopen please provide test cases, examples and an update to the documentation.

PR #211, opened 2013-12-16
Fix for #527 - 3.4: Queryable not configurable

diegomigliavacca:
http://tracker.deegree.org/deegree-services/ticket/527

It's possible to set a layer as queryable or not queryable setting the attribute "enabled" on the "l:FeatureInfo" element. This configuration was not working properly.


stephanr:
I think this one should be superseded by #519 which fixed the issue on a wider scope.

diegomigliavacca:
yes, the first part of the PR #519 fixes this issue as well.
I can close this pull request.

PR #215, opened 2014-01-14
Fix for #609 - 3.3: Tile raster layers are not rendered in a good way

diegomigliavacca:
http://tracker.deegree.org/deegree-services/ticket/609


tfr42:
Thank you for your contribution. Your pull request has been discussed by the TMC. Please see the following remarks from the TMC members:

The TMC would be happy if can find the time to address these shortcomings as described above.

tfr42:
Is anyone needing this bug fix? Can someone post the issue which this PR is trying to fix? We need input from the author.

diegomigliavacca:
This pull request has been around for long time. I close it and I will possibly submit it again later, whenever I will find a fix which doesn't break the integration tests.

PR #388, opened 2014-08-19
Improve deegree webservices distribution - fixes issue #336

### LINKED ISSUE #336: Drop maven module org.deegree:deegree-tomcat in favor of using dependency to Tomcat in deegree-webservices
Actual the deegree module org.deegree:deegree-tomcat contains a copy of Apache Tomcat/6.0.29. This plain vanilla Tomcat comes with some deegree toppings which add no additional functionality besides checking the JVM and setting JAVA_HOME.  
The expected resulting context is that this module is dropped and in org.deegree:deegree-webservices a dependency to a more recent version of Apache Tomcat 7.x or even 8.x is defined. Additionally the assembly has to be extended.

tfr42:
Fixes #336 - dropped deegree-misc/deegree-tomcat module at all and replaced the distribution bundle of Apache Tomcat with deegree-webservices as root web app as simple maven assembly in deegree-services/deegree-webservices-tomcat-bundle


tfr42:
Replaced by #477

PR #407, opened 2014-09-10
ssl

dstenger:
Added ssl support to OwsHttpClient.


tfr42:
Thank you for your contribution. Unfortunately, after carefully reviewing your contribution we, the TMC members, came to the conclusion not to merge it into the deegree code base. Of course this is not the final word. If you think, that we did wrong then please re-initiate the Decision Making Process by making a proposal and start a discussion on the mailing list. Perhaps we can reach a consensus then. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #430, opened 2014-10-08
Removed catch (Throwable) in OracleSqlDialect to improve error handling and logging

dstenger:
Caught throwables were removed (and throw of new IllegalArgumentExceptions without message) to improve the exception handling (now, the "original" exception is thrown instead of a new IllegalArgumentException).
In addition, the logging will be more detailed as log messages can now contain the "original" exception instead of the new IllegalArgumentException.


stephanr:
The TMC discussed that it would be better to keep the IllegalArgumentException with the orginal Stack Trace.

So i will change the code and prepare a pull request for the next TMC Meeting

tfr42:
PR #545 is replaces this one.

PR #442, opened 2014-10-27
Extend tiling api to be able to transform raster images

dstenger:
The tiling API was extended to be able to to transform raster images. Thie pull request does NOT include a raster transformer, so an error is returned if request crs does not match the tile crs (same behaviour in the current implementation, just the error type changed from invalidCrs (in my opinion a wrong type as a user can configure other crs) to illegalRequest)

There are two general problems:
- The source tiles have a different crs and envelope than the request. How can the correct tiles be choosen?
- Raster image transformations are not 100% accurate. Due to that fact, there are small gaps between the tiles if the transformation is done for each tile itself.

To tackle the first problem, the request envelope is transformed to the crs of the first coordinate system known to the TileLayer. Then all tiles are retrieved with the transformed envelope from the tile data set having the first crs (canges in TileLayer). Later, the Java2DTileRenderer checks whether the crs of the tiles and the request crs match. Otherwise a transformation is executed (or error is thrown if there was no transformer injected into tile renderer).

To tackle the second problem, the image is created in the tile crs first and finally the whole image is transformed to the query crs (almost all changes in Java2DTileRenderer). A precondition for this mechanism was to change the API of the TileRenderer: All tiles are given to the renderer instead of just one single.

Finally, the Java2DTileRenderer can now get an ImageTransformer (new interface) via constructor injection. This transformer is used to transform the raster image. If null is injected (status of this pull request), an error is thrown if an user tries to transform tile datsources (same behaviour as in the current implementation; see first passage).


tfr42:
Thank you for your contribution. Your pull request has been discussed by the TMC today. Please see the following remarks from the TMC members:
 * How do I actually use this functionality via a TileStore config?
 * Do all TileStore types support this?

dstenger:
* How do I actually use this functionality via a TileStore config?
There is no extra configuration necessary. If the storage CRS of the TileStore differs from the requested CRS, transformation is automatically activated if ImageTransformer has been injected. Otherwise, an exception is thrown if CRS differ from each other (also current behaviour).
 * Do all TileStore types support this?
The feature was tested and implemented for GeoTIFFTileStore. Other TileStores have not been tested, but they should work as well.

tfr42:
Dear deegree users,
If someone of you would like this PR merged and want this as an official part of deegree, then please act now! Either rebase and fix the build or post a comment. Otherwise we will close this PR soon.

tfr42:

PR #443, opened 2014-10-27
Created geotools transformer and injected it into the tile renderer

dstenger:
This pull request is on the basis of https://github.com/deegree/deegree3/pull/442 . The other pull request should be dealt with first as this one just adds the geotools transformer.

A geotools transformer (class: GeotoolsRasterTransformer; implements ImageTransformer) was implemented to transform raster images from one envelope to another envelope. As the name of the class already implicates, the transformer uses the geotools library (which is currently not present in the deegree nexus).

There are several reasons why geotools was chosen to do the transformation of a raster image:
- Raster tranformation is a complex feature which is tackled by geotools quite good (why should I implement my own RasterTransformer?).
- The API of raster transformation is simple in geotools.
- There is a raster transformer in the coverage API of deegree, but its implementation has several drawbacks (complex API, not documented well) and did not work well with the tiling API (axis order was a big problem).
- The quality of the result image is better with geotools than with the raster transformer of the coverage API.

Finally, the geotools transformer was injected into the Java2DTileRenderer (in class DefaultRenderContext) and is used to transform tile datasources.


tfr42:
Thank you for your contribution. Your pull request has been discussed by the TMC today. Please see the following remarks from the TMC members:

 * How is this actually used in a TileStore configuration?
 * Which types of TileStores support this?

dstenger:
* How do I actually use this functionality via a TileStore config?
There is no extra configuration necessary. If the storage CRS of the TileStore differs from the requested CRS, transformation is automatically activated if ImageTransformer has been injected. Otherwise, an exception is thrown if CRS differ from each other (also current behaviour).
 * Do all TileStore types support this?
The feature was tested and implemented for GeoTIFFTileStore. Other TileStores have not been tested, but they should work as well.

tfr42:
Thanks @dstenger for answering. Is it possible that you can also add a short description to the deegree manual in chap. "7. Tile stores"? We could ask for community support as well.

tfr42:
Dear deegree users,
If someone of you would like this PR merged and want this as an official part of deegree, then please act now! Either rebase and fix the build or post a comment. Otherwise we will close this PR soon.

tfr42:

PR #460, opened 2014-11-27
Enhance handling of nodata values in coverages

wanhoff:
Until now, pyramid configurations did not provide a way to set a nodata value. That results in unexpected images, if the bounding box is not completely covered by the provided image or i.e. the image is cut with the elevation parameter.
Pixels not covered by the image or cut out are assigned a value of 0, but that is not neccessarily the nodata value of the image. So it may be rendered in an unexpected colour.

This pull request provides the possibility to configure the value set for nodata pixels. If it is set to the appropriate value (like nodata or background colour in style configuration), it will be rendered as expected.


tfr42:
Thank you for your contribution. Unfortunately, it cannot be merged automatically. It cannot be considered for inclusion, before this has been fixed. Please rebase it against the target branch.

tfr42:

PR #466, opened 2014-12-22
Fix NPE in XPath evaluation

MrSnyder:
Happened on empty primitive-valued elements, e.g. when evaluating aixm:name/text() on

```xml
   <aixm:name xsi:nil="true"/>
```


tfr42:

PR #469, opened 2015-01-10
Add SVG support to WMS

copierrj:
This pull adds support for SVG to the WMS. Based on a proof of concept implementation by @erikorbons.

It required a change in the GDAL layer implementation that might hurt performance in some cases.

### DIFF
--- a/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java
+++ b/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java
@@ -57,12 +57,6 @@ import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.io.OutputStream;
 
-import org.deegree.rendering.r2d.Java2DRasterRenderer;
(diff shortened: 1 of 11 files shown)


tfr42:
Short statement to add about the performance impact, in which scenario does this occur and  how can this made configurable.

copierrj:
Regarding the performance issue:

Currently the GDAL layer replaces the image used by the render context (as render target) instead of painting the GDAL images on top of the existing image. This results in a performance increase. 

This behavior results in the wrong image when multiple layers are requested (what usually isn't happening when using GDAL).

In the SVG render context we cannot use this approach as we do not have an image as render target. Therefore I removed this performance 'hack' during implementation of the SVG support.

Possible solution: a lazy image render context that doesn't allocate an image until one is needed. See: IDgis@469318e01c117acc2a67d8b8c470e06aa757219b

copierrj:
I've tested the previously mention 'lazy image render context' approach with the GDAL layer and after some tweaking concluded that the resulting behavior is as expected: when only a single GDAL layer is rendered, the image obtained from GDAL is returned to the client without any additional processing.

tfr42:

PR #483, opened 2015-01-22
SQLFeatureStore: Respect configured feature resolver in BLOB mode

MrSnyder:
Prior to this patch, for features queried by id in BLOB mode, the resolver was always set to a new instance of FeatureStoreGMLIdResolver.


tfr42:

PR #494, opened 2015-02-06
Bugfix/3.4 renderer clipping and crash fix

stephanr:
This Bugfix fixes the Polygon rendering of clipped Polygons and also extends the logic of clipping for circumstances to be more robust.

I also moved the clipping 	decision methods to the GeometryClipper to be easier readable.

Currently if a stroke contains a dasharray an a logical defect geometry (for example one point a million meter away from the rest of the points) the jdk renderer will either consume a extrem large ammount of time or in the worst case the jdk will crash.

As this situation is extreamly annoying i also created a testcase to prevent the clipper from missing this situation.
(So for example if the patch would be reversed the JUnit Testcase will catch the timeout and fail)

### DIFF
--- a/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/GeometryClipper.java
+++ b/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/GeometryClipper.java
@@ -103,6 +103,10 @@ class GeometryClipper {
                 if ( jtsOrig == jtsClipped ) {
                     return geom;
                 }
+                if ( isInvertedOrientation( jtsOrig ) ) {
+                    return clippedGeometry;
+                }
+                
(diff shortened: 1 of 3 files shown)


stephanr:
The deegree 3.3 Line is not affacted because all Geometries will be clipped [1].

[1] https://github.com/deegree/deegree3/blob/3.3-master/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/Java2DRenderer.java

stephanr:
On the TMC meeting we discussed that, 
fixin this pull request in the 3.4 line will depend on if this fix can be made wihtout altering a generally handling of feature orientation.
I will analyse if this can be done in a reasonable matter of time to keep it as bugfix for 3.4 or will change the milestone to 3.5.

stephanr:
I had rebased the pull request and applied a workaround for CW/CCW oriented polygons so they are not reorientated anymore. 
If the build will pass travis the milestone can be keept.

PR #496, opened 2015-02-07
Remove GetMapFormat enumeration

copierrj:
Suggest to remove the GetMapFormat from the wms configuration xml schema.

Rationale: 
- prevent inconsistencies between Java code and this xml schema.
- the wms shouldn't deal directly with supported image formats because output format handling is provided by the OutputFormatProvider. At the moment we've only one implementation of this OutputFormatProvider but this could (should?) be an interesting extension point.

### DIFF
--- a/deegree-services/deegree-services-wms/src/main/java/org/deegree/services/wms/controller/WMSController.java
+++ b/deegree-services/deegree-services-wms/src/main/java/org/deegree/services/wms/controller/WMSController.java
@@ -735,15 +735,20 @@ public class WMSController extends AbstractOWS {
         return featureInfoManager;
     }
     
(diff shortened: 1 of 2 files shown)


tfr42:
Do we need to discuss this new extension point in more details?

copierrj:
During the TMC meeting we discussed the wish to not remove the enumeration from the xml schema but only lift the restrictions in order to maintain Eclipse code completion.

copierrj:
Not ready yet: needs rebase + implementation of suggested fix.

copierrj:
For consideration: replicate the 'feel' of the GetFeatureInfo configuration option.

tfr42:

PR #498, opened 2015-02-09
Switch travis to new container-based infrastructure and enable caching

stephanr:
This Pull Request tests the new container-based infrastructure of travis which enables public repositories to use a faster an cacheing enabled build environment [1]. 

[1] http://docs.travis-ci.com/user/workers/container-based-infrastructure
[2] http://docs.travis-ci.com/user/caching

Before this pull Request can be used, the travis build result has to be checked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+# cache for public repo on container-based infrastructure see:
+# http://docs.travis-ci.com/user/workers/container-based-infrastructure
+# http://docs.travis-ci.com/user/caching
+#
+sudo: false
 language: java
 jdk:
   - openjdk7
@@ -5,3 +10,6 @@ jdk:
 install: "travis_retry mvn install -Pintegration-tests -DskipTests=true -B -V"
 script: "travis_retry mvn verify -Pintegration-tests -B -V"
 env: MAVEN_OPTS="-XX:MaxPermSize=128m"
+cache:
+  directories:
+    - $HOME/.m2/repository/


MrSnyder:
Wow, this looks very promising.

MrSnyder:
Travis succeed in about 35 minutes. I assume that this was not using caching yet. Initiated a restart, fingers crossed :-)

stephanr:
The point that the build passes after 35 / 38 min would also a nice benefit as i had seen many builds that failed after +/- 50 mins of build time.

MrSnyder:
Looks even better now. 28~29 mins :-)

tfr42:
Great improvement!

stephanr:
After compareing the logs of the two builds, the one with cache is about 0,5 MiB smaler than without. 
So we would have more space before the build gets killed by travis at 4 MiB [1].

One little imperfection has the log at least, that the cache could not be updated with the second run:
> The command "travis_retry mvn verify -Pintegration-tests -B -V" exited with 0.
store build cache
> change detected: /home/travis/.m2/repository/org/deegree/deegree/3.4-pre17-SNAPSHOT/_remote.repositories
> changes detected, packing new archive
> uploading archive
> failed to upload cache
> curl: (55) SSL_write() returned SYSCALL, errno = 104
> 
> Done. Your build exited with 0. 

This might only be related to the fact that the pull request introduces the cache to to travis configuration. 
And on the other hand the maven logic very robust and the cache could also be cleared manually  by the tmc via web on the travis page.
20150209_buid471_caches

So for now, i would like to see this change merged into deegree as soon as possible so that we could see if the builds become more stable.

[1] Example of failed build ending
> The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).
> 
>  The build has been terminated.

PR #503, opened 2015-02-12
Bugfix/3.4 fix JDK 6 compatibility

stephanr:
In Pull Request #466 a JDK 6 the Collections.emptyIterator() was used which is not available in JDK6 see [1].
The class also used the Jaxen.EMPTY_ITERATOR constant, so i switched to this and i also harmonized the use of the constant.

[1] http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#emptyIterator%28%29


tfr42:

Issue #516, opened 2015-03-20
Disable GetFeatureInfo for single Layer not Working as expected 3.4


stephanr:
If the configuration of a theme has only layer which are disabled by configuration (see [1]) the Layer still is `queryable="1"` in WMS/GetCapabilties and could be queried by Request.

In this situation it is not relevant if the element ist disabled by setting:
```xml
<l:LayerOptions>
	<l:FeatureInfo enabled="false" />
</l:LayerOptions>
```
or by specifieing the backward compatible setting of:
```xml
<l:LayerOptions>
	<l:FeatureInfoRadius>0</l:FeatureInfoRadius>
</l:LayerOptions>
```

A demo workspace is available and can be provided if needed, please let me know.

[1] http://buildserver.deegree.org/job/deegree-3.4-master/ws/deegree-services/deegree-webservices-handbook/target/html/layers.html#rendering-options

stephanr:
This issue was fixed by #518

Issue #513, opened 2015-03-19
build issues because of deegree3-*.deegree.org not available


stephanr:
I had some problems running local builds with test and integraton-test enabled because some tests refferes to deegree3-demo.deegree.org and deegree3-testing.deegree.org, which are not available at the moment.
This maybe also apply to travis build (not yet tested)

I had seen at least one place where these services (utha-demo) are used directly and also in workspaces setup for remote-stores.

I thinks this may be relative to issue #488 and the migration of the demos.

@tfr42 Were the a different set of demos at deegree3-demo/deegree3-testing ?

A quick workaround test rewriting deegree3-demo and deegree3-testing to demo.deegree.org (via changed hosts entry and a apache with proxy) restored the failed test which directly reffered to deegree3-*.deegree.org.

If the demos where equal a possible solution would be to change the remaining dns C-NAMES and enhance the virtual-host configuration of the plaza.deegree.org.

stephanr:
The Problem occoured on both lines 3.3 and 3.4

tfr42:
> Were the a different set of demos at deegree3-demo/deegree3-testing?
Not that I am aware of. The target environment plaza.deegree.org shall host all the services of demo.deegree.org. The proposed fix to configure Apache httpd to rewrite deegree3-demo and deegree3-testing to demo.deegree.org sounds ok.

@wanhoff @NielsWalen Can you help out?

Long term goal is to get the integration tests self-contained and to avoid any access of external services. This would require major rework on the integration tests.

stephanr:
The rewriting thing was a local test for me to check if the tests failed because of the remote demo or because of other issues.

For plaza.deegree.org i think it should suffice to add deeegree3-demo/deegree-testing to the vhost alias (This could be done by me), but the DNS C-NAME alias has to be changed by someone else.

NielsWalen:
Both are added with pointer to plaza.deegree.org. Thanks for the hints.

stephanr:
Ok, i had altered the Apache config on plaza.deegree.org so deegree3-demo/deegree3-testing/demo are now served by the same tomcat.

So we will see if that will fix the build problems.

stephanr:
The failed test count reduced to one failed WMTS test which i suspect that the demo may have changed here, so i have to dig in here :-/
````
testCapabilitiesPyramidLayerNoFeatureInfoFormats(org.deegree.services.wmts.WmtsIT)  Time elapsed: 0.108 sec  <<< FAILURE!
[...]

PR #518, opened 2015-03-20
3.4 bugfix featureinfo queryable

stephanr:
This Pull Requests fixes the the issue #516 by
 * correcting the loading of the layer options (correct the `enabled=false`)
 * enhancement of handling of the three states of the feature info radius
  * < 0 ==> default (as specified in service endpoint)
  * = 0 ==> disabled
  * > 0 ==> user specified radius
 * fixing the `LyerMetadateMerger` to make a Layer in WMS/GetCapabilties only `queryable=0` if all layers in the sub structure are not queryable


stephanr:
The builds failed because of problems with the deegree3-demo/deegree3-testing wich are referenced by the tests, so #513 has to be resolved before this test ca be rerun

PR #524, opened 2015-03-31
Fixed bug in sld_body request parameter

dstenger:
The request parameter "SLD_BODY" did not work as the style/filter/layer maps in RequestBase were never filled when parsing the sld. This fix enables the filling of those maps.

This pull request was also created for the master (#523).

### DIFF
--- a/deegree-core/deegree-core-protocol/deegree-protocol-wms/src/main/java/org/deegree/protocol/wms/ops/RequestBase.java
+++ b/deegree-core/deegree-core-protocol/deegree-protocol-wms/src/main/java/org/deegree/protocol/wms/ops/RequestBase.java
@@ -149,10 +149,10 @@ public abstract class RequestBase {
                 }
             }
 
-            this.layers.clear();
             this.styles.clear();
-            
             this.filters = new LinkedList<OperatorFilter>();
+            
+            LinkedList<LayerRef> tmpLayers = new LinkedList<LayerRef>();
             
             // to get the order right, in case it's different from the SLD order
             for ( LayerRef lRef : layers ) {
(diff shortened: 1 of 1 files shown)


MrSnyder:
Thank you for your contribution. However, we found that it is not ready yet for inclusion:
 * We need more information when this issue occurs  
 * What is the context?
  * Please provide information on the current behavior and the expected behavior.
  * Can you please provide a request which does not work without that fix and works with that fix 

The TMC would be happy if you can find the time to address these shortcomings.

tfr42:
Moved to 3.3.16

segoerke:
This issue occurs in every deegree webservices version between 3.2 and this fix.
The current behaviour is, that requests such as:
http://yourfavoritesite.com/WMS? 
   REQUEST=GetMap& 
   BBOX=0.0,0.0,1.0,1.0& 
SLD=http%3A%2F%2Fmyclientsite.com%2FmySLD.xml&
   WIDTH=400& 
   HEIGHT=400& 
   FORMAT=PNG 

or

http://yourfavoritesite.com/WMS? 
   REQUEST=GetMap& 
   BBOX=0.0,0.0,1.0,1.0& 
   SLD_BODY=%3C%3Fxml+version%3D%221.0%22+encoding%3D%22UTF-8%22%3F%3E%3CStyled
LayerDescriptor+version%3D%221.1.0%22%3E%3CNamedLayer%3E%3CName%3ERivers%3C%2FN
ame%3E%3CNamedStyle%3E%3CName%3ECenterLine%3C%2FName%3E%3C%2FNamedStyle%3E%3C%2
FNamedLayer%3E%3CNamedLayer%3E%3CName%3ERoads%3C%2FName%3E%3CNamedStyle%3E%3CNa
me%3ECenterLine%3C%2FName%3E%3C%2FNamedStyle%3E%3C%2FNamedLayer%3E%3CNamedLayer
%3E%3CName%3EHouses%3C%2FName%3E%3CNamedStyle%3E%3CName%3EOutline%3C%2FName%3E%
3C%2FNamedStyle%3E%3C%2FNamedLayer%3E%3C%2FStyledLayerDescriptor%3E 
   WIDTH=400& 
   HEIGHT=400& 
   FORMAT=PNG 

are processed while ignoring the sld parameters like SLD_BODY.
This means, that using the sld parameters given by the OGC SLD specification did never work in the mentioned versions.
The final result of that fix is, that now the sld parameters are used to style the features of the requested layers instead of doing nothing.

copierrj:
See: #606

PR #525, opened 2015-04-09
Improve XPath support for TimeInstant/TimePeriod objects

MrSnyder:
This patch improves the support for accessing XML nodes inside of TimeInstant/TimePeriod objects.

Consider that the following GML example has been parsed into deegree's object model:

```
<TimePeriod gml:id="p1" xmlns="http://www.opengis.net/gml/3.2" xmlns:gml="http://www.opengis.net/gml/3.2">
  <beginPosition>2001-05-23</beginPosition>
  <endPosition indeterminatePosition="unknown" />
</TimePeriod>
```

Before the patch, it was not possible to access the value of the "indeterminatePosition" attribute. After the patch xpaths that target this attribute are properly evaluated, e.g.

```
gml:TimePeriod/gml:endPosition/@indeterminatePosition
```


tfr42:

PR #526, opened 2015-04-09
SQLFeatureStore: Support all GMLObject types in relational mode feature building

MrSnyder:
Prior to this patch, a GML object other than a feature or a geometry would be recreated as GenericXMLElement by a relational-mode SQLFeatureStore. After this patch, such objects are transformed to the respective GMLObject subtype (e.g. TimeSlice or TimeObject). This is important for in-memory filtering on these kinds of objects.


tfr42:
Besides the lax resource handling in FeatureBuilderRelational.recreatePropertyFromGml() everything is fine. If you could check if the close() statements can move to a finally block would be nice.

tfr42:

PR #527, opened 2015-04-09
SQLFeatureStore: Allow arbitrary SQL expressions in primitive particle mappings

MrSnyder:
Prior to this patch, it was possible to use database column names or constants in the mapping attribute of primitive particles, e.g.:
```
<Primitive path="@nilReason" mapping="nil_reason"/"
```
or
```
<Primitive path="@nilReason" mapping="'unknown'"/"
```
After this patch, it's also possible to use arbitrary SQL expressions in the mapping of a primitive particle. Here's an example that maps a gml:validTime property. In this case, it either contains a TimeInstant, a TimePeriod or has attribute 'nilReason' set to 'unknown'. The attribute is set in case both the gml_validtime_gml_timeinstant and the gml_validtime_gml_timeperiod columns are NULL:

```
<Complex path="gml:validTime">
  <Primitive path="@nilReason" mapping="(CASE WHEN $0.gml_validtime_gml_timeinstant IS NULL AND $0.gml_validtime_gml_timeperiod_begin IS NULL AND $0.gml_validtime_gml_timeperiod_end IS NULL THEN 'inapplicable' ELSE NULL END)" type="text" />
  <Complex path="gml:TimeInstant">
    <Complex path="gml:timePosition">
      <Primitive path="text()" mapping="gml_validtime_gml_timeinstant" type="dateTime" />
    </Complex>
  </Complex>
  <Complex path="gml:TimePeriod">
    <Complex path="gml:beginPosition">
      <Primitive path="@indeterminatePosition" mapping="CASE WHEN gml_validtime_gml_timeperiod_begin IS NULL AND gml_validtime_gml_timeperiod_end IS NOT NULL THEN 'unknown' ELSE NULL END" type="text" />
      <Primitive path="text()" mapping="gml_validtime_gml_timeperiod_begin" type="dateTime" />
    </Complex>
    <Complex path="gml:endPosition">
      <Primitive path="@indeterminatePosition" mapping="CASE WHEN $0.gml_validtime_gml_timeperiod_end IS NULL AND $0.gml_validtime_gml_timeperiod_begin IS NOT NULL THEN 'unknown' ELSE NULL END" type="text" />
      <Primitive path="text()" mapping="gml_validtime_gml_timeperiod_end" type="dateTime" />
    </Complex>
  </Complex>
</Complex>
```
The SQL expression for determining the value of the indeterminatePosition attribute is:
```
CASE WHEN $0.gml_validtime_gml_timeperiod_begin IS NULL AND $0.gml_validtime_gml_timeperiod_end IS NOT NULL THEN 'unknown' ELSE NULL END
```
Note that column names in the expression need to be aliased with "$0.": When this expression is passed through to the SQL database, $0 is replaced with the correct table alias used in the SELECT column list or in the WhereBuilder.

### DIFF
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/feature/property/GenericProperty.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/feature/property/GenericProperty.java
@@ -169,6 +169,12 @@ public class GenericProperty implements Property {
         this.children = Collections.singletonList( value );
     }
 
+    public GenericProperty( PropertyType declaration, QName name, TypedObjectNode value, Map<QName,PrimitiveValue> attrs ) {
+        this( declaration, name, value );
+        this.attrs = attrs;
+        this.children = Collections.singletonList( value );
+    }    
+    
     @Override
     public QName getName() {
         return name;
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/GMLStreamReader.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/GMLStreamReader.java
@@ -145,6 +145,8 @@ public class GMLStreamReader {
 
     private GMLDictionaryReader dictReader;
 
+    private boolean laxMode;
+
     /**
      * Creates a new {@link GMLStreamReader} instance.
      * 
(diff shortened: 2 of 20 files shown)


tfr42:

PR #529, opened 2015-04-10
3.4 Coverage Layer Heap Enhancement / Rendering Performance Enhancement

stephanr:
This pull requests makes two optimizations to the CoverageLayerData to achieve a faster rendering if a large coverages or a high number of coverages are asigned to a single wms GetMap Operation.

This is achieved by:
 * only rendering raster which intersects with the requested bbox
 * not holding the transformed request for the current CoverageLayerData in memory until the requests ends

For Example: 
A WMS-GetMap request of 2000x2000 pixel against 200 coverages holds a peak of ~ 3000 MB (4 byter per pixel) at the last processed CoverageLayerData in the heap instead of ~ 32 MB after the patch.


tfr42:

PR #536, opened 2015-04-16
Enhancement allowing to set more than one MetadataURL for each Layer/FeatureType

dstenger:
Currently, just one metadataUrl is set for each Layer/FeatureType in the capabilities document of the WMS and WFS, even if multiple metadataUrls are configured for a Layer/FeatureType (if this is the case, just the last one is written in the capabilities document).
This fix allows to configure multiple metadataUrls for a single Layer/FeatureType and returns them in the capabilities response. The fix has been tested for WMS 1.3.0, WFS 1.1.0 and WFS 2.0.0.

In addition, two general adjustments were done:
 * Previously, the WFS just returned FeatureTypes with metadataUrls in the capabilities document if metadataUrls were set. This behaviour was changed and now always all FeatureTypes are returned (even if metadataUrls are set).
 * Previously, the metadataUrls of the first layer were also set in top level layer (e.g. root layer) in the capabilities document of the WMS. This behaviour was changed and now top level layer do not get the metadataUrls of the first child layer anymore.


MrSnyder:
Thank you for your contribution. Your pull request has been discussed by the TMC.

However, we found that it is not ready yet for inclusion. Our concern is the first general adjustment (behaviour of WFS was changed and now always reports all feature types in GetCapabilities response). This reverts behaviour that has been introduced in 3.4 *on purpose*. What is this purpose? Several people asked (e.g. on the mailing list) for the possibility to limit the list of reported feature types in a schemas-based WFS, especially in the context of INSPIRE. An INSPIRE theme may contain more than hundred feature types, but usually just a few are actually used in an INSPIRE Download Service. Therefore, we would like to ask to remove this part of the change. 

segoerke:
Where is this changed behaviour documented? I was not able to find any hint about this configuration possibility within the documentation master. So I've added a new issue (https://github.com/deegree/deegree3/issues/555) for that, as this missing information now leads into additional efforts for us...

segoerke:
I think the related pull request is: https://github.com/deegree/deegree3/pull/304
Unfortunately there is NO documentation about this behaviour besides the pull request...
Additionally I think, this behaviour is not what users want by default. Maybe there is metadata for only one FeatureType avaiable but all FeatureTypes shall be provided through the service... IMHO this is a critical shortcoming which leads into the situation that people are forced to provide metadata for everything or for nothing...

Why does the TMC accept such pull requests (no hints in documentation, forced behaviour by default, no configuration options) but rejects others with the same issues like https://github.com/deegree/deegree3/pull/542.
Does it belong to if the pull request comes from a TMC member if it is accepted while failing the acceptance criteria?

dstenger:
Relevant commit was removed.
Following general adjustment was reverted by that:
 * Previously, the WFS just returned FeatureTypes with metadataUrls in the capabilities document if metadataUrls were set. This behaviour was changed and now always all FeatureTypes are returned (even if metadataUrls are set).

PR #537, opened 2015-04-16
Enhancement allowing to set more than one MetadataURL for each Layer/FeatureType

dstenger:
Currently, just one metadataUrl is set for each Layer/FeatureType in the capabilities document of the WMS and WFS, even if multiple metadataUrls are configured for a Layer/FeatureType (if this is the case, just the last one is written in the capabilities document).
This fix allows to configure multiple metadataUrls for a single Layer/FeatureType and returns them in the capabilities response. The fix has been tested for WMS 1.3.0, WMS 1.1.1, WFS 1.1.0 and WFS 2.0.0.

In addition, two general adjustments were done:
 * Previously, the WFS just returned FeatureTypes with metadataUrls in the capabilities document if metadataUrls were set. This behaviour was changed and now always all FeatureTypes are returned (even if metadataUrls are set).
 * Previously, the metadataUrls of the first layer were also set in top level layer (e.g. root layer) in the capabilities document of the WMS. This behaviour was changed and now top level layer do not get the metadataUrls of the first child layer anymore.

This pull request was also created for the master (#536).

### DIFF
--- a/deegree-services/deegree-services-commons/src/main/java/org/deegree/services/metadata/OWSMetadataProvider.java
+++ b/deegree-services/deegree-services-commons/src/main/java/org/deegree/services/metadata/OWSMetadataProvider.java
@@ -106,4 +106,13 @@ public interface OWSMetadataProvider extends Resource {
      */
     DatasetMetadata getDatasetMetadata( QName name );
 
+    /**
+     * Returns a list of data metadata for the specified dataset.
+     * 
+     * @param name
+     *            for layers, a qname with only a local name is used, for feature types its qname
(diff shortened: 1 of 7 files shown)


tfr42:
Please see the comment by the TMC in #536

lgoltz:
Relevant commit was removed. Following general adjustment was reverted by that:

 * Previously, the WFS just returned FeatureTypes with metadataUrls in the capabilities document if metadataUrls were set. This behaviour was changed and now always all FeatureTypes are returned (even if metadataUrls are set).

PR #542, opened 2015-04-28
Use X-Forwarded-Host from header as HTTP GET and POST DCP URL if available

dstenger:
This enhancement enables following functionality:
If an incoming request contains the header "X-Forwarded-Host", the DCP URLs in the capabilities document are replaced by the value of the header. If the header is not present, the old mechanism is used.
The enhancement allows, for example, that a deegree service which can be accessed by multiple URLs returns the correct DCP URLs if the "X-Forwarded-Host" header is set.

### DIFF
--- a/deegree-services/deegree-services-commons/pom.xml
+++ b/deegree-services/deegree-services-commons/pom.xml
@@ -91,7 +91,15 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-    </dependency>     
+    </dependency>   
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito</artifactId>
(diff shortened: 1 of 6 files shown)


lgoltz:
Is the hard coded header name fine or is a configuration option required?

tfr42:
Thank you for your contribution. Your pull request has been discussed by the TMC. Please see the following remarks from the TMC members:

 * Can you please add some explanation into the deegree web service handbook in [chapter 5.7.1](http://download.deegree.org/documentation/3.3.14/html/webservices.html#reported-urls) how this works.

However, we found that it is not ready yet for inclusion. The TMC would be happy if can find the time to address these shortcomings.

dstenger:
Documentation (chapter 5.7.1) was added to the pull request.

In addition, a few adjustments were done. Now, the X-Forwarded-Host header just replaces the host of the DCP URL (and not the protocol and port). Those two values are taken from the incoming request instead.
Further, the header X-Forwarded-Proto and X-Forwarded-Port are regarded, if the X-Forwarded-Host header is set. Those header are used (if set) as protocol and/or port instead of the values of the incoming request.

PR #543, opened 2015-04-29
Fixed SQL-Statement generated for resultType="hits" where column names are not unique

lgoltz:
This bug fix allows !GetFeature-requests with resultType="hits" in a SQLFeatureStore configuration with multiple tables where two or more tables have columns with the same name.
The alias of the root table is added in the select clause, to avoid SQLExceptions.


tfr42:

PR #545, opened 2015-04-29
Removed catch (Throwable) in OracleSqlDialect to improve error handling and logging on (rework)

stephanr:
The caught throwable in the OracleGeometryConverter are now thrown as Stack of a new IllegalArgument Exception.
This was made to:
 * Arrive the same goal as #430 to allow a better logging including original Exception information
 * Keep the same Code workflow (throw IllegalArgumentException)


tfr42:

PR #561, opened 2015-05-11
WFS: Omit already exported features in additional objects section

MrSnyder:
Prior to this patch, it was possible that circular references caused the AdditionalObjects section of a WFS 2.0.0 response to contain wfs:member elements with xlinked content, e.g.

```
<wfs:FeatureCollection...>
...
  <wfs:additionalObjects>
    ...
    <wfs:member xlink:href="#uuid.bedfbe9d-cb04-4915-9e72-3f09b2940e7b" />
   ...
  </wfs:additionalObjects>
<wfs:FeatureCollection/>
```
This is not schema-valid and also doesn't make sense: The additionalObjects section is supposed to contain "real" features, not references.


vog:
Travis CI failed but there was to error message at the end. What should we do about that?

PR #572, opened 2015-06-09
Fixes bug in ConnectionProviderManager: Removed deregistration of drivers from DriverManager

lgoltz:
In the scenario described as follows an SQLException occurs. Two deegree-webservice instances with jdbc connection(A and B) are running in one tomcat. A is undeployed or stopped via tomcat manager, B is requested where the processing of the requests requires a database connection.

Exception:
```
org.springframework.transaction.CannotCreateTransactionException: 
Could not open JDBC Connection for transaction; 
nested exception is java.sql.SQLException: No suitable driver found for ....
 ```
The fix removes the part of the shutdown method in the ConnectionProviderManager where all drivers are registered from java.sql.DriverManager via reflections. As far as I understand this should remove all drivers loaded with the module classloader of the workspace. But the field  *registeredDrivers* of the java.sql.DriverManager contains instances of DriverInfo (an inner class of the DriverManager). The check if the classloader is same as the module classloader comes never true, but the null check does (as drivers are loaded in bootstrap class loader, which can be represented by null). This leads to the removing of **all** drivers overall deegree-webservices instances.

### DIFF
--- a/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/jdbc/DriverWrapper.java
+++ b/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/jdbc/DriverWrapper.java
@@ -94,6 +94,11 @@ public class DriverWrapper implements Driver {
         return d.jdbcCompliant();
     }
 
+    @Override
+    public String toString() {
+        return "Wrapped: " + d.toString();
+    }
+
     public Logger getParentLogger()
                             throws SQLFeatureNotSupportedException {
         Logger logger = null;
--- a/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/jdbc/DriverWrapper.java
+++ b/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/jdbc/DriverWrapper.java
@@ -55,7 +55,7 @@ import java.util.logging.Logger;
  */
 public class DriverWrapper implements Driver {
 
-    private Driver d;
+    private final Driver d;
 
     public DriverWrapper( Driver d ) {
         this.d = d;
@@ -97,6 +97,10 @@ public class DriverWrapper implements Driver {
     @Override
     public String toString() {
         return "Wrapped: " + d.toString();
+    }
+
+    public Driver getWrappedDriver() {
+        return d;
     }
 
     public Logger getParentLogger()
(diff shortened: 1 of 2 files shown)


tfr42:
Thanks @lgoltz for sharing that finding and the bug fix. 
We have to make here a general statement about handling resources within deegree and JDBC connections in particular when deploying multiple web apps with deegree into a servlet container such as Apache Tomcat.
The suspicious code de-registers *all* JDBC driver classes from the [DriverManager] (https://docs.oracle.com/javase/7/docs/api/java/sql/DriverManager.html) even then when loaded from the parent class loader or other locations. The method ```ConnectionProviderManager.shutdown()``` does this trick. Why was this implemented? deegree allows to put extension modules into the ```modules``` directory which might include also a JDBC driver (see [Java modules and libraries] (../blob/master/deegree-services/deegree-webservices-handbook/src/main/sphinx/javamodules.rst)). But this is not the recommend way (anymore)! Since we are already using Java SE 7 with JDBC 4.0 and deegree 3.4 supports [javax.sql.DataSource] (https://docs.oracle.com/javase/7/docs/api/javax/sql/DataSource.html) there are better places for a JDBC driver JAR and better ways to retrieve a JDBC connection.

First question: Where to place a JDBC driver JAR?

A JDBC driver shall be loaded by the class loader (or one of it's child class loader) which loaded the ```DriverManager```. JDBC drivers register themselves in the JVM-wide singleton ```DriverManager``` which is shared by all web apps. The ```DriverManager``` class will attempt to load the driver classes referenced in the "jdbc.drivers" system property. This allows a user to customize the JDBC Drivers used by their applications.' (from the JavaDoc). In Apache Tomcat place the JDBC driver into the ```lib```folder or add the path to the JAR file to the CLASSPATH environment variable. Read more in the [Apache Tomcat Class Loader HOW-TO](http://tomcat.apache.org/tomcat-8.0-doc/class-loader-howto.html) or check the documentation of your preferred Java EE servlet container for more information.
This will also prevent memory-leaks when re-deploying a web app since both DriverManager and JDBC driver class come from non-web app classloaders.
Keep in mind:
Never put different versions of the same JDBC driver into the class path!

Second questions: How to configure a JDBC connection (DataSource)?

We encourage users of deegree to use the new feature introduced with PR #233 supporting [DataSources] (../blob/master/deegree-services/deegree-webservices-handbook/src/main/sphinx/serverconnections.rst). 

Third question: Multiple deegree web application within one single servlet container?

Yes, this is possible and even recommended. But keep in mind to deploy the same JDBC driver only once and not with each web app again .

To wrap up:
We have to add this information to the web services handbook and the code should reject loading JDBC drivers from the ```modules``` directory.

tfr42:
Duplicate of #580

PR #576, opened 2015-06-19
SQLFeatureStore: Improve support for filter-based queries without feature type (e.g. by gml:identifier)

MrSnyder:
This pull request drastically improves the performance of the FeatureStoreGmlIdentifierResolver.

Some WFS scenarios (e.g. in AIXM) use gml:identifier-based references (xlinks in the GML representation) instead of gml:id-based references. When such a reference has to be resolved, every feature type has to be considered, because the target feature type is usually not known.

Prior to this patch, the FeatureStoreGmlIdentifierResolver performed a Filter-based query for each feature type. This can be very slow, for example when the SQLFeatureStore has to perform hundreds of individual SELECTs just to resolve a single reference. For some configurations (most notably the BLOB mode), a single SELECT would be sufficient, but there's currently no way to express a filter-based query that targets more than a single feature type.

### DIFF
--- a/deegree-datastores/deegree-featurestores/deegree-featurestore-commons/src/main/java/org/deegree/feature/persistence/FeatureStoreGmlIdentifierResolver.java
+++ b/deegree-datastores/deegree-featurestores/deegree-featurestore-commons/src/main/java/org/deegree/feature/persistence/FeatureStoreGmlIdentifierResolver.java
@@ -35,8 +35,8 @@
  ----------------------------------------------------------------------------*/
 package org.deegree.feature.persistence;
 
-import java.util.ArrayList;
-import java.util.List;
+import static org.deegree.filter.MatchAction.ANY;
+
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


vog:
Travis CI failed, but I can't see any error message. What should we do about that?

copierrj:
Is this part of the SQLFeatureStore covered by any test? Can we sure this is not breaking anything?

tfr42:
@vog Can you please answer @copierrj question? How did you test this enhancement?

vog:
@tfr42, @copierrj: Unfortunately, I can't answer this question in detail. I hope that @MrSnyder will provide more insight into this topic.

MrSnyder:
I just added a description of the idea behind the changes to the pull request.

We tested the patch in AIXM-based configurations that use the FeatureStoreGmlIdentifierResolver and tested the correct behavior in practice.

MrSnyder:
I will provide a slightly different pull request including tests. Closing this one.

Issue #531, opened 2015-04-14
ISOMetadataStore (SQL): Setup tables script fails on PostgreSQL 9.3/PostGIS 2.1


MrSnyder:
I am setting up an ISOMetadataStore (SQL) using deegree 3.4-pre17 on Ubuntu 14.04 LTS (packaged PostgreSQL/PostGIS version). I run into the following issue when creating the database tables using "Setup tables":

"Error: ERROR: function addgeometrycolumn(unknown, unknown, unknown, unknown, unknown, unknown) is not unique Hint: Could not choose a best candidate function. You might need to add explicit type casts. Position: 8"

I analyzed the SQL script and found the problem in line 75:

"SELECT AddGeometryColumn('public','idxtb_main','bbox','-1','GEOMETRY','2');"

Here's what works for me:

"SELECT AddGeometryColumn('public','idxtb_main','bbox',0,'GEOMETRY',2);"

I think there are two problems:
* SRID and dimension should be given as numbers, not strings
* Since some time, the official "undefined" SRID is 0, not -1

segoerke:
> Since some time, the official "undefined" SRID is 0, not -1
That's the case since PostGIS 2.0

PR #582, opened 2015-07-01
Fix for #531-3.3: CSW error when creating database tables

diegomigliavacca:
This pull request fixes the issue #531 in Deegree 3.3.


tfr42:
A hint to the handbook should be added.

copierrj:
Alternative solution: change the sql to use a conditional expression to differentiate the srs (0 for PostGIS 2 and -1 for PostGIS 1.5)

tfr42:
The best approach is to use a conditional expression to differenciate between postgis 1.4, 1.5 and 2.0
in the Postgres dialect.

tfr42:
Replaced by #658

PR #615, opened 2015-10-27
Implemented Temporal Operators After, Before und During for WFS 2.0.0

dstenger:
Implemented the temporal operators "After", "Before" and "During" for WFS 2.0.0.

Temporal operators can be used to filter GetFeature requests.

The support of those operators is described in the capabilities and the constraints "ImplementsMinTemporalFilter" and "ImplementsTemporalFilter" are set to true.


tfr42:
@MrSnyder your feedback is appreciated. Can you please have a look on the implementation?

tfr42:
Further testing conformance might be needed. Currently the OGC CITE TestEngine for WFS 2.0.0 does not cover the temporal conformance class.

PR #622, opened 2015-11-05
Enhanced support for layer attributes in WMS 1.3.0, 1.1.1

lgoltz:
This pull requests adds the export of layer attributes to the WMS Capabilities.
- Default value of "0" for "noSubsets", "fixedHeight" and "fixedWidth" always appears in capabilities.
- Attribute "queryable" always appears in capabilities (default value is "0").
- Exception with code "LayerNotQueryable" is thrown, if layer with "queryable" of "0" is requested.
- A new configuration option was added to the layer configuration to allow the user to set the value of the "opaque" attribute. Value always appears in the capabilities (default value is "0").
- Value of "1" is set for "cascaded", if remote OWS is used. For superior layer the highest value of the child layer is selected. Attribute "cascaded" always appears in capabilities (default value is "0") .


dstenger:
The build error occurs as the basic configuration of layers was changed (base.xsd). The schema is imported to other layer configurations (coverage, feature, gdal, remotewms, tile). The Problem is that those configurations point to the schema available on http://schemas.deegree.org/:
```xml
<import namespace="http://www.deegree.org/metadata/description" schemaLocation="http://schemas.deegree.org/commons/description/3.1.0/description.xsd" />
<import namespace="http://www.deegree.org/metadata/spatial" schemaLocation="http://schemas.deegree.org/commons/spatialmetadata/3.1.0/spatialmetadata.xsd" />
<import namespace="http://www.deegree.org/layers/base" schemaLocation="http://schemas.deegree.org/layers
```
This schema does not contain the changes. In module deegree-wmts-tests various layer artifacts are included. In all of them the class LayerOptionsType can be found, but the changes are just present in deegree-core-layer, which causes the error.

A possible solution is to update the schema on http://schemas.deegree.org/.

copierrj:
Travis is failing because of #649

tfr42:
Check is needed if the updated base.xsd has been uploaded by Jenkins job after release completed.

tfr42:
It requires to upload the XSD file before release job is executed.

tfr42:
Version of the file needs to be updated to 3.4, see #650.

tfr42:
Also check that the XSD files are packaged into the JAR files and that the files are resolved from the JAR files.

tfr42:
This PR is accepted for 3.4. Currently we can't merge this PR due to the fact that it is blocked by the issue with the schema upload and update #650.

Issue #612, opened 2015-10-06
loss of performance in 3.4-pre18 SQLFeatureStore


lgoltz:
Commit https://github.com/deegree/deegree3/commit/b25892700f21a47b388265284b8a8370a3335e53 leads to a loss of performance in deegree-3.4-pre18. The commit was part of the pull requests #526  and #527. 
The same WFS GetFeature request can be answered with 3.4-pre17 in ~0,3 seconds, in 3.4-pre18 it needs ~27,7 seconds (the same workspace is used). 
The method causing the problem is #recreatePropertyFromGml() in org.deegree.feature.persistence.sql.rules.FeatureBuilderRelational.

MrSnyder:
That's bad.

The goal of the mentioned commits is to ensure that mapped GML objects are represented using the correct classes of deegree's TypeObjectNode hierarchy. My use case is the correct in-memory filtering of gml:TimeInstant and gml:TimePeriod elements for an SQLFeatureStore configuration for AIXM documents. For that, such elements need to be represented as TimeInstant-/TimePeriod-objects.

I am aware that the taken approach is somewhat hacky, but I didn't realize that it would affect any relevant use cases. Can you identify the relevant mappings in your SQLFeatureStore configuration?

The idea of the patch is to simply use deegree's GML parser for any in order to transform GenericXmlElement instances into TimeInstant-/TimePeriod-objects. This way, I was able to avoid the huge code changes that "native" relational building of such objects would introduce ( for now).

A possible quick fix could be to put a check  after line 328 in FeatureBuilderRelational. Here, we could try to determine if this is a property that contains a gml:TimeInstant or a gml:TimePeriod object. Otherwise, we should skip the call to recreatePropertyFromGml. However, it should first be clarified for which kind of element mapping the regression occurs.

lgoltz:
Here is the FeatureTypeMapping of the requested FeatureType:
```xml
<FeatureTypeMapping name="dog:Hauskoordinaten" table="haus">
        <FIDMapping prefix="BB_HK_">
            <Column name="gid" type="integer"/>
        </FIDMapping>
        <Primitive path="iso19112:geographicIdentifier" mapping="geographicidentifier"/>
        <Complex path="iso19112:geographicExtent">
            <Complex path="gmd:EX_BoundingPolygon">
                <Complex path="gmd:polygon">
                    <Geometry path="." mapping="geographicextent">
[...]

PR #623, opened 2015-11-05
Performance improvement for SQLFeatureStore fixing #612

dstenger:
This pull requests fixes the Bug reported in #612 by adding an instanceof check for TimeObject.


tfr42:
Fixes #612

PR #631, opened 2015-11-16
Ignore Oracle SDO LRS (Linear Referencing System)

copierrj:
This pull request proposes to ignore any LRS information while converting an Oracle SDO (= JDBC STRUCT) to a deegree Geometry.

Currently trying to convert such a STRUCT into a deegree Geometry results in an exception explaining that LRS handling is not supported by deegree.

Unfortunately it is no trivial matter (as far as I know) to properly map an Oracle SDO with LRS information onto the deegree geometry model. Nevertheless, deegree should be able to process the non LRS part of the data. Therefore I would like to suggest to ignore the LRS dimension and treat the remaining dimensions as usual.

The change is implemented by complementing existing code in order to prevent accidentally breaking something.


tfr42:
@stephanr please check this until the next TMC meeting.

stephanr:
The TMC decided that it would helpful to test the code change against a seperated setup.
I will do this in preparation for the next TMC meeting.

stephanr:
I had succesfully tested the PR against our LRS dataset and everything worked as intendet.
So i think this PR is ready to get included.

PR #638, opened 2015-12-01
Fixed exception code if unknown FeatureType is inserted by WFS 2.0.0 transaction

dstenger:
An InvalidValues exception should be thrown by WFS 2.0.0  transaction if unknown FeatureType is inserted:
 * WFS 2.0.0 Specification, Table 3 - WFS exception codes, P.16:
  * InvalidValue: "A Transaction (see Clause 15) has attempted to insert or change the value of a data component in a way that violates the schema of the feature."

This fix changes the exception code to "InvalidValues" if an unknown FeatureType is inserted by WFS 2.0.0 transaction.


tfr42:
The official OGC CITE tests have been used to verify the expected behaviour.

PR #661, opened 2016-01-05
Enhances WMS to throw an exception if the requested CRS is not configured

lgoltz:
Currently the WMS throws an exception if the requested CRS  in GetMap and GetFeature requests is not supported by deegree. This pull requests changes this behavior to throw an InvalidSRS/InvalidCRS exception if the requested CRS is not configured (not declared in the capabilities).
To avoid breaking of clients expecting the old behavior this is configurable. By default deegree checks only if the CRS is not supported. To limit this to the configured CRS, the service configuration must be extended: 

```
<wms:CrsCheckStrict>true</wms:CrsCheckStrict>
```


tfr42:
PR is OK. But for the RC line we accept bug fixes only. This is why we have put it onto milestone 3.5. It might be included in 3.4.1.

copierrj:
The TMC discussed this pull request which resulted in a more in depth discussion related to the proposed xml schema change. We decided to not yet merge this PR and continue the discussion in the next TMC meeting. See for details: #953

PR #665, opened 2016-01-12
Integration tests for SQLFeatureStore

MrSnyder:
This pull request adds a new module (deegree-featurestore-sql-tests) for SQLFeatureStore integration tests. The goal is to provide an automated means for checking regressions (the CITE test configurations are based on the MemoryFeatureStore).

Currently, the module contains a tiny framework and a basic test case based on AIXM 5.1 application schemas. We tried to design it in a simple way to make it easy to add tests for other schemas.

Please have a look at class SQLFeatureStoreTestCase (and AIXMQueryIT) to learn how to use it. It's based on XMLTestCase from XMLUnit to make easy to check generated features against an the expected result (in GML).

It requires a PostgreSQL database with PostGIS extension. On Travis CI, such an instance is  provided by the build environment automatically [1]. Again, it should be easy to add tests based on other databases (e.g. Oracle).

In order to run the tests locally, a database with standard credentials must be available on localhost (port: 5432, user: postgres, password: postgres). Each test creates the database "deegree-test" and drops it afterwards.

Note that the actual changes introduced by the patch are minimal. No changes outside the new module, the large number of files is due to the GML application schemas used for the AIXM test.

[1] https://docs.travis-ci.com/user/using-postgresql/


tfr42:
Please split this PR into separate PR which might include the fix for the ```SQLFeatureStore``` the other with the integration tests. Furthermore the hard-coded values for the postgresql JDBC connection shall be externalized into a properties for or so. Also, we would like to have a profile which indicates that this is an integration test to be executed on travis-ci. We do favor platform and environment independent integration tests. Using a test infrastructure based on docker must be supported.

MrSnyder:
Torsten, thank you for your comments. Some thoughts:

* Splitting the Pull request: I did include the bugfix on purpose, as the integration tests triggers it. I could split it easily, but it would be hard to have a unit test for it then. And it would create a dependency between the PRs. What do you think?
* Properties file for the JDBC connection. I see your point. Where shall I put it so people don't miss it? Any clever idea how to cope with the fact that Travis will require this file to exist?
* Using profile "integration-tests": This is already the case, as the whole module subtree "deegree-tests" only runs for this profile. See root POM.
* Where shall I put the developer documentation? Shall I create a new Wiki page or integrate it somewhere?

MrSnyder:
Ok. I removed the fix and adapted the tests accordingly, so the PR does not introduce any changes outside the new module.

Remaining concerns:

* Properties file for the JDBC connection. I see your point, and we also discussed it at m-click. We decided to go for convention over configuration here, as in the time of Vagrant, it may be easier to bring up a PostGIS db for the right connection parameters than dealing with config parameters at all.  I can offer to provide a Vagrantfile for a PostGIS setup that uses the same parameters as the Travis PostGIS instance does. In case the TMC does not like that, I would like to ask for a hint where to put the properties file, so people don't miss it. Also we need cope with the fact that Travis will require this file to exist.
* Using profile "integration-tests": This is already the case, as the whole module subtree "deegree-tests" only runs for this profile. See root POM.
* Where shall I put the developer documentation? I did write some JavaDoc and updated the description in the ticket. Is that enough or shall I create a new Wiki page or integrate it somewhere?

tfr42:
Postponed to the next TMC meeting. Needs further analysis.

tfr42:
The TMC has decided today that this PR won't most likely be merged into the code base. We still consider the idea behind this PR as very useful for the deegree project but we are in favor of following a different approach. Here are our considerations:
- the overloaded maven build is a big burden already, instead of squeezing another integration test into the maven build we are in favor to have a additional validation step such as a system test or  a functional acceptance test step. The pipeline approach on CI systems is up-and-comming.   
- the infrastructure for setting up the SQLFeatureStore for different database backends should not rely on pre-installed software or services in hosted environments such as travis-ci. Here we do favor a build infrastructure based on docker where the infrastructure is defined as code. Top notch! 

So, you are invited to help us to take this one step further and finally integrate the tests into deegree. Contributions are welcome!

tfr42:

PR #673, opened 2016-01-29
Fixed bug in GmlGetFeatureHandler with WFS 2.0.0 and GetFeatureById

sNiklasch:
When the WFS 2.0.0 request is GetFeatureById (example below), the xmlStream had no start element to insert the namespace with prebindNamespaces, what went into an error.
So the "prebindNamespaces" was moved into the if-blocks.

Example request:

```xml
<wfs:GetFeature xmlns:wfs="http://www.opengis.net/wfs/2.0" count="10" service="WFS"
	                startIndex="0"
	                version="2.0.0">
	<wfs:StoredQuery id="urn:ogc:def:query:OGC-WFS::GetFeatureById">
		<wfs:Parameter name="id">EXAMPLE.123456</wfs:Parameter>
	</wfs:StoredQuery>
</wfs:GetFeature>
```


sNiklasch:
Duplicate of #620 because of rebase miss, closed.

PR #681, opened 2016-02-24
[WIP]3.4 Allow custom particle converter in table driven mode

stephanr:
Currently CustomParticleConverter can be configured for a SQLFeaturesStore regardless the operation mode of the SQLFeatureStore.

The Current Implementation only used the CustomParticleConverter if the SQLFeatureStore is in schema driven mode and ignores it otherwise.

As this is already fully available logic, it only has to be passed to activate it also for the table driven mode. 
The [documentation](http://download.deegree.org/documentation/3.4-RC1/html/featurestores.html#customizing-the-mapping-between-columns-and-properties) only lists the posibility to use this extension point but has no details for it.

It gets even more usefull if the query can be accesed from the thread wich is curently querying the data.

### DIFF
--- a/deegree-datastores/deegree-featurestores/deegree-featurestore-sql/src/main/java/org/deegree/feature/persistence/sql/SQLFeatureStore.java
+++ b/deegree-datastores/deegree-featurestores/deegree-featurestore-sql/src/main/java/org/deegree/feature/persistence/sql/SQLFeatureStore.java
@@ -160,6 +160,8 @@ public class SQLFeatureStore implements FeatureStore {
 
     private static final int DEFAULT_CACHE_SIZE = 10000;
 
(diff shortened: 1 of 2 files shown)


copierrj:
Question: is it possible to give a converter access to the query object without using a ThreadLocal?

tfr42:
Discussed during TMC meeting:
- use a single point where Thread-local variables are bind to a thread (e.g. OGCFrontController servlet)
- API shall provide an extension point to add this additional logic

tfr42:
Postponed to the next TMC meeting.

stephanr:
I have reviewed the situation and decided to change this pull to replace the thread local with a public static property as (open) extension point.

I will try to suply this solutaion as soon as possible and then remove the [WIP]

stephanr:
I close this PR, because it has been waiting for a long time and there are no planned resources for it at the moment.

PR #682, opened 2016-02-27
SQLFeatureStore: Fix filtering on gml:identifier in BLOB mode

MrSnyder:
Before this patch, queries that filter based on the gml:identifier property would not match any features. The patches fixes that so the expected feature is returned. Test case included.

Depends on #665.


copierrj:
The TMC reviewed this PR and decided that in order to consider it for merging it should be rebased first. Although not technically required by git, we would like to this PR to be rebuild / retested by travis using the pom.xml and travis configuration currently in master.

copierrj:
Reminder to TMC: infrastructure changes as outlined in #851 should be finalized before merging this issue. Otherwise we might not be able to do release builds anymore!

vog:
This pull request depends on #665, so it makes no sense to rebase it before #665 is resolved.

tfr42:

PR #683, opened 2016-02-28
SQLFeatureStore: Improve support for queries without typenames

MrSnyder:
Prior to this patch, it was generally not possible to use the SQLFeatureStore#query methods without specifying the feature type name. The sole exception have been queries with filters on the feature id. For operator filters, the only way to find matching features without knowing the feature type beforehand was to perform individual queries for each feature type

After this patch, it is possible to create Query objects without the name of the feature type and use such queries on the SQLFeatureStore both in BLOB and in relational mapping mode.

Besides improving the convenience of the API, the real benefit is that in BLOB mode, the query is performed much more efficiently. Instead of n SELECTs, only a single SELECT will be performed. For relational mode, a CombinedFeatureResultSet is used that iterates over all mapped feature types, so the API contract is met and streaming is kept intact.

Notable use cases for this API improvement:
- WFS 2.0 queries that specify an abstract feature type
- Finding features by gml:identifier (tests included)

Depends on #682.


tfr42:
Thank you for your contribution. Unfortunately, the TMC members came to the conclusion not to merge it into the deegree code base since the PR get stuck (see [PR marked for deletion](https://github.com/deegree/deegree3/wiki/Working-with-pull-requests#pull-requests-marked-for-deletion)). Of course this is not the final word. If you think, that we did wrong then please re-initiate the Decision Making Process by opening an updated pull request and taking the feedback from the TMC into consideration. Perhaps we can then merge that new pull request. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #684, opened 2016-02-28
Improve FeatureStoreGMLIdentifierResolver performance

MrSnyder:
Prior to this patch, resolving a single xlink using the FeatureStoreGMLIdentifierResolver (e.g. in AIXM configurations) was performed using individual queries for each mapped feature type. Depending on the configuration this could mean dozens or hundreds of SELECTs.

After this patch, the FeatureStoreGMLIdentifierResolver uses the facility introduced by #683, so only a single query without typename is issued against the feature store.

Depends on #683 (needs to be pulled first).


tfr42:
Thank you for your contribution. Unfortunately, the TMC members came to the conclusion not to merge it into the deegree code base since the PR get stuck (see [PR marked for deletion](https://github.com/deegree/deegree3/wiki/Working-with-pull-requests#pull-requests-marked-for-deletion)). Of course this is not the final word. If you think, that we did wrong then please re-initiate the Decision Making Process by opening an updated pull request and taking the feedback from the TMC into consideration. Perhaps we can then merge that new pull request. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #685, opened 2016-02-28
Add support for partial SQL and partial post filtering in WhereBuilder

MrSnyder:
Prior to this patch, the WhereBuilder would either map a complete filter to a WHERE clause or (in case some parts of the filter are unmappable) use an empty WHERE clause and offer the complete filter for in-memory filtering.

A shortcoming of this approach is that a single unmappable part of a filter means that all objects will be be fetched from the database and will be filtered in memory. Depending on the structure of the filter, this can be improved, so the WHERE-clause can still be used to narrow the result set.

After this patch, the WhereBuilder supports partial SQL and partial in-memory filtering for a common use case: If the filter is based on a root AND operator, the filter will be split into a WHERE clause and a post filter part. The WHERE clause contains all mappable parts and the post filter the remaining parts.

Tests included.

### DIFF
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/filter/Filters.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/filter/Filters.java
@@ -35,7 +35,9 @@
  ----------------------------------------------------------------------------*/
 package org.deegree.filter;
 
+import static java.util.Arrays.asList;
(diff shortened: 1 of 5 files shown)


copierrj:
The TMC discussed this pull requests and we like this approach. We are however somewhat hesitant to merge this right now because we very much like this new behavior to be opt-in instead to the  new default behavior. Adding a configuration setting for this would be a great improvement to this pull request.

Issue #511, opened 2015-03-10
WFS-T update problem when using 3D-coordinates


larseggan:
I have set up a WFS-Transaction server with deegree 3.3.8 (SQL Feature store)
I want  to use 3D-coordinates:
wfs:Insert is OK with xyz, but wfs:Update failes.
The error message from wfs:Update:
<?xml version='1.0' encoding='UTF-8'?>
<ows:ExceptionReport 
    xmlns:ows="http://www.opengis.net/ows/1.1" 
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" version="2.0.0">
    <ows:Exception exceptionCode="NoApplicableCode">
        ows:ExceptionTextGeometry is invalid. Dimensionality of coordinates (=3) does not match dimensionality of CRS 'urn:ogc:def:crs:EPSG::25832' (=2)./ows:ExceptionText
    /ows:Exception
/ows:ExceptionReport

I have used EPSG:25832 (ETRS89 / UTM zone 32N), but the correct EPSG code is a compound crs EPSG:6172 (ETRS89 / UTM zone 32 + NN54 height) or EPSG:5972 (ETRS89 / UTM zone 32 + NN2000 height).
Is there any way of getting these crs into deegree?
Regards,
Lars

tfr42:
Unfortunately the handbook is missing this information. But please have a look on the pull requests #174 and #238 as an example. There are 3 configuration files you might have to touch: ``crs-definitions.xml``, ``datum-definitions.xml``, and  ``projection-definitions.xml``.

larseggan:
I followed the following description: http://wiki.deegree.org/deegreeWiki/deegree3/WorkspaceConfiguration/CRSConfiguration
And then added a new crs to crs\deegree\d3\config\crs-definitions.xml:
<crs:CompoundCRS>
                <crs:Id>epsg:6172</crs:Id>        
                <crs:Id>urn:ogc:def:crs:epsg::6172</crs:Id>
                <crs:Name>ETRS89 / UTM zone 32N +NN54 height</crs:Name>
                <crs:Version>2015-03-11</crs:Version>
                <crs:Description>Inserted by hand.</crs:Description>
               <crs:AreaOfUse>6.0,57.93,12.0,65.76</crs:AreaOfUse>
[...]

PR #693, opened 2016-05-02
Fix to allow inserting 3d

jarped:
Added check for crs-dimension before writing.


jarped:
Should fix https://github.com/deegree/deegree3/issues/511 - unverified

PR #722, opened 2016-07-05
Enhanced RemoteWMS Layer to use XSLT files for GetFeatureInfo requests

dstenger:
Previously, when GetFeatureInfo was requested on a RemoteWMS Layer, a default parser for the returned XML of the remote service was used. As the returned XML often has different structures depending on the implementation of the remote service, the default parser already considers different structures (e.g. ESRI has its own structure) but it is, of course, limited to the in deegree implemented cases.

This pull request introduces the possibility to set a XSLT file in the RemoteWMS Layer which is used to configure the parser for the GFI response of the remote service. By that, deegree users can now configure the parser making it possible to use GFI responses of remote services which cannot be parsed by the default parser or control the structure of the GML.

The element

```
<XSLTFile targetGmlVersion="GML_2">name2gml.xslt</XSLTFile>
```

is used to activate a XSLT file for GetFeatureInfo requests to a remote layer.
The target GML version must also be configured.

Full configuration example of the new option:

```
  <Layer>
    <!-- Layer name of the original service -->
    <OriginalName>Name</OriginalName>
    <!-- deegree configuration identifier to be use e.g. in theme configuration file -->
    <Name>name</Name>
    <XSLTFile targetGmlVersion="GML_2">name2gml.xslt</XSLTFile>
  </Layer>
```


deegree-core/deegree-core-featureinfo/src/main/java/org/deegree/featureinfo/parsing/DefaultFeatureInfoParser.java
```diff
@@ -0,0 +1,239 @@
...
+public class DefaultFeatureInfoParser implements FeatureInfoParser {
+
+    private static final Logger LOG = getLogger( DefaultFeatureInfoParser.class );
+
+    private static final XMLInputFactory XML_FACTORY = XMLInputFactory.newInstance();
+    
+    @Override
+    public FeatureCollection parseAsFeatureCollection( InputStream inputStream, String csvLayerNames )
+                            throws XMLStreamException {
+        XMLStreamReader xmlReader = XML_FACTORY.createXMLStreamReader( inputStream );
```
stephanr, line 94: As a remark, we have a security optimized way of getting a safe at https://github.com/deegree/deegree3/blob/master/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/xml/stax/XMLInputFactoryUtils.java#L60

PR #725, opened 2016-07-12
Enhanced ReferenceResolvingException to be multilingual

lgoltz:
Previously the ReferenceResolvingException was a hard coded message in english. With this pull requests a german message was provided by using i18n messages from ResourceBundle.


copierrj:
The TMC is  (in principle) in favor of i18n efforts. However: if we are going to endorse this we would expect that (eventually) every string is translated. Not just this one.

copierrj:
The TMC re-evaluated the need for i18n within deegree. Although we are by no means against i18n in general we would like it to be a all or nothing solution. Translating just a single error message into another language doesn't seem to be the right approach.

PR #734, opened 2016-08-11
Enhanced WFS for support of CreateStoredQueries and DropStoredQuery

lgoltz:
The operations CreateStoredQueries and DropStoredQuery was implemented. With this enhancement deegree supports the conformance class _A.1.15 Manage stored queries_ . The implementation discriminates between three different kinds of stored queries:
- fixed - Fixed stored queries are pre configured, like GetFeatureById for WFS 2.0.
- configured - Configured stored queries are configured by the deegree adminstrator.
- managed - Managed stored queries are inserted via  CreateStoredQueries.

Fixed and configured stored queries can not be dropped by  DropStoredQuery request. If an   attempt was made to drop a fixed or configured stored query an InvalidParameterException is thrown.

The managed stored queries are stored in the subfolder _storedqueries/managed_ in the  workspace. The directory must exist to enable the feature.

If an already existing StoredQuery is attempted to be created a second time, a DuplicateStoredQueryIdValue exception is thrown (Status code 409).

If a StoredQuery is created in an unsupported language, an exception is thrown.


copierrj:
Enabling this feature by creating a folder in the workspace is somewhat unconventional. Please consider to make this an explicit configuration option.

PR #754, opened 2016-08-23
3.4 remove wps jasper reports

stephanr:
On the FOSS4G Bonn CopdeSprint we (@tfr42, @dstenger, @lgoltz ) discussed to drop in 3.4 the wps extension for the jasperreports.
This was initiated to remove the (old) dependecies of the version of jaspersoft and its dependencies.

At last, as far as i know, the jaspersoft-wps-modules were not used in the webservices bundle.

This pull request moves the workspace and the modules to the uncoupled.

**TODO:**
- [x] Check if the handbook containes references to the jrxml module, which also have to be removed


stephanr:
I just checked, the handbook does not contain a reference to the jrxml process provider.

stephanr:
After rebase and close/open the pull request builds fine

Issue #674, opened 2016-02-05
Create WPS client ProcessExecution from status location URL


awst-baum:
Is it possible to create a WPS client ProcessExecution (org.deegree.protocol.wps.client.process.ProcessExecution) from a status location URL as returned by org.deegree.protocol.wps.client.process.ProcessExecution.getStatusLocation()? I can't find a method to do it - but it would be really useful if I could just store the status URL for a long-running process and later on restore the ProcessExecution from it, without the need to keep the original ProcessExecution object alive.

I guess the content for that method would be similar to org.deegree.protocol.wps.client.process.ProcessExecution.getState() but would return a new ProcessExecution. A quick fix would be to derive my own CustomProcessExecution and implement it, but sadly all necessary fields and the constructor are private.

Is there already such a method? And if not, could it be considered as a future feature?

awst-baum:
My solution would be to add something like this to org.deegree.protocol.wps.client.WPSClient.WPSClient:

`    /***
     * Creates an ExecutionResponse object (contains the process execution status and outputs)
     * from the URL to an wps:ExecuteResponse document.
     * @param url URL of a processExecution's wps:ExecuteResponse document (obtain via 
     * execution.getStatusLocation()).
     * @return ExecutionResponse object containing the process execution status and outputs. 
     * @throws OWSExceptionReport
     * @throws IOException
     * @throws XMLStreamException
     */
    public ExecutionResponse getExecutionResponse(URL url) throws OWSExceptionReport, IOException, XMLStreamException {
        LOG.debug( "Polling response document from status location: " + url );
        XMLInputFactory inFactory = XMLInputFactory.newInstance();
        InputStream is = url.openStream();
        XMLStreamReader xmlReader = inFactory.createXMLStreamReader( is );
        XMLStreamUtils.nextElement( xmlReader );
        if ( OWSExceptionReader.isExceptionReport( xmlReader.getName() ) ) {
            throw OWSExceptionReader.parseExceptionReport( xmlReader );
        }
        ExecuteResponse100Reader reader = new ExecuteResponse100Reader( xmlReader );
        return reader.parse100();
    }`
It could be re-used in org.deegree.protocol.wps.client.process.ProcessExecution.getState().

I'm happy to hear why this is a stupid idea ;-)

tfr42:
Please post this as a feature request and if possible a related PR. See our [guidelines for working with PR](https://github.com/deegree/deegree3/wiki/Working-with-pull-requests) for further information.

PR #760, opened 2016-08-29
Method to create WPS client ExecutionResponse from status location URL

awst-baum:
This implements the behaviour requested in issue #674:
Two (convenience) methods are added that generate an ExecutionResponse object (with execution status, percent complete, etc.) from a status location URL or stream.

To avoid code duplication, these methods are then used by ProcessExecution to parse execution responses as well.

The WPS client integration test is expanded to use the new methods.


copierrj:
Travis currently has a problem building this PR. Rebasing would likely fix this.

tfr42:
@awst-baum  Thank you for your contribution. Unfortunately, it cannot be merged automatically. It cannot be considered for inclusion, before this has been fixed. Please rebase it against the target branch.

Issue #696, opened 2016-05-04
International characters in QName (3.4 RC1)


jarped:
In 3.4 RC1, deegree is unabe to load a featurestore which is set up to serve an application-schema that contains international characters (æøå in norwegian).

The same featurestore works in 3.3.18

Errormessage:

> Error near line 16, column 83: cvc-datatype-valid.1.2.1: 'app:RpOmrÃÂ¥de' is not a valid value for 'QName'.
> Error near line 16, column 83: cvc-attribute.3: The value 'app:RpOmrÃÂ¥de' of attribute 'name' on element 'FeatureTypeMapping' is not valid with respect to its type, 'QName'.

The feature is named app:RpOmråde

tfr42:
@jarped  Can you please post the complete feature store configuration file.

jarped:
Here's one that runs on the same schema, should give the same problems:
https://raw.githubusercontent.com/jarped/deegree.featurestores/reguleringsplaner/reguleringsplan_20160615.xml

tfr42:
@jarped  SNAPSHOT Version for testing is available. Please get in contact with us if you can do some testing.

jarped:
I've been testing some. I think the system default is being used. In my case this is windows-1252.

**java -version**
>java version "1.7.0_80"
Java(TM) SE Runtime Environment (build 1.7.0_80-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.80-b11, mixed mode)

Running on Windows 10.

If I open an already existing featureStore (encoded id UTF-8), special characters seem to be interpreted as windows-1252.

I've also tried to copy config (https://raw.githubusercontent.com/jarped/deegree.featurestores/reguleringsplaner/reguleringsplan_20160615.xml) into the deegree-editor. When I then try to save it crashes:

> Caused by: java.io.CharConversionException: Invalid UTF-8 start byte 0xf8 (at char #1367, byte #-1)
        at com.ctc.wstx.io.UTF8Reader.reportInvalidInitial(UTF8Reader.java:303)
        at com.ctc.wstx.io.UTF8Reader.read(UTF8Reader.java:189)
        at com.ctc.wstx.io.ReaderSource.readInto(ReaderSource.java:87)
        at com.ctc.wstx.io.BranchingReaderSource.readInto(BranchingReaderSource.java:57)
        at com.ctc.wstx.sr.StreamScanner.loadMore(StreamScanner.java:991)

dstenger:
[...]

PR #789, opened 2016-12-14
Fixed broken characters caused by the deegree console

dstenger:
Previously, special characters in configuration files were broken after editing and saving with the deegree console. This PR sets UTF-8 explicitly as character encoding to avoid broken characters.

PR for 3.4: #790


tfr42:
Fixes #696

Issue #772, opened 2016-09-15
Avoid usage of default namespace in test resources


tfr42:
As in #739 and #764 described.

awst-baum:
Please check out whether #792 is what you expect :)

PR #792, opened 2016-12-21
Avoid usage of default namespace in curve.xml

awst-baum:
Fixing bug #772: Avoid usage of default namespace in curve.xml


awst-baum:
@tfr42 Any chance of moving forward with this pull request? It's not a big change, you asked for it yourself, and without it optional WPS client tests fail, which is a pain in the neck when unit testing changes.

PR #799, opened 2017-01-03
move from urn to url in test

jarped:


jarped:
Closing due to resolved in mysterious ways

PR #801, opened 2017-01-10
Stable version: WPS client: Use URIs in complex outputs as well as inputs

awst-baum:
This extends #795: now the WPS client outputs also use URI instead of URL.


tfr42:
PR for master branch is missing. PR need to be prepared.

tfr42:
Changes are OK.

PR #806, opened 2017-01-17
Allow characters from Latin-1 supplement in fieldNames 

jarped:
sans multiplication and division
https://unicode-table.com/en/blocks/latin-1-supplement/


copierrj:
@jarped Could you please verify if the PR #812 does fixes your issue?

PR #814, opened 2017-02-03
Find correct featureStore when inserting - 3.3

jarped:


jarped:
I see some problems regarding featureCollections.

tfr42:
#815 is PR for 3.4

tfr42:
- [ ] The PR #815 needs rebase and hence we merge 3.3 and 3.4 in a row this PR is put on hold
- [ ] Our concern about this PR is that it handles not all cases when dealing with multiple FeatureStores. We would like to better understand what is the problem which is should be solved by this PR?

copierrj:
This pull request seems stuck for quite a long time. Therefore the TMC decided to close this pull request.

Also note: the 3.3 branch is no longer supported and no subsequent releases are to be expected.

PR #815, opened 2017-02-03
Find correct featureStore when inserting - 3.4

jarped:


tfr42:
- [ ] Update  to documentation is missing.

- [ ] Not all cases are handled, we consider this as incomplete. Please get in contact with us.

tfr42:
Thank you for your contribution. Unfortunately, the TMC members came to the conclusion not to merge it into the deegree code base since the PR get stuck (see [PR marked for deletion](https://github.com/deegree/deegree3/wiki/Working-with-pull-requests#pull-requests-marked-for-deletion)). Of course this is not the final word. If you think, that we did wrong then please re-initiate the Decision Making Process by opening an updated pull request and taking the feedback from the TMC into consideration. Perhaps we can then merge that new pull request. Anyhow, we do appreciate your commitment to open source and to deegree in particular.

PR #819, opened 2017-02-28
Enhanced WMS GetMap request by the vendor specific parameters FILTERPROPERTY and FILTERVALUE to filter the underlying data source

lgoltz:
This PR adds the two vendor specific parameters FILTERPROPERTY and FILTERVALUE to the deegree WMS. The parameters allows the client to filter the underlying datasource of the requested layers.
FILTERPROPERTY contains the name of the property to filter for. It is no qualified, so only the local name is considered. FILTERVALUE contains a list of values to filter for, multiple values must be comma separated. All features with the property matching one of the value are rendered.

It the layer does not base on feature datasource or does not provide a property with the given name, the filter is ignored. Same if one of the parameter is missing or the value empty.

### DIFF
--- a/deegree-core/deegree-core-layer/src/main/java/org/deegree/layer/LayerQuery.java
+++ b/deegree-core/deegree-core-layer/src/main/java/org/deegree/layer/LayerQuery.java
@@ -36,11 +36,16 @@
 package org.deegree.layer;
 
 import static java.lang.Integer.parseInt;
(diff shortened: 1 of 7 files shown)


copierrj:
During the TMC meeting a question was raised whether we want to extend deegree with vendor specific parameters as this makes swapping out deegree for some other OGC implementation more challenging.

PR #687, opened 2016-03-08
Upgrade Apache Commons Collection to v3.2.2

ProgramMax:
Version 3.2.1 has a CVSS 10.0 vulnerability. That's the worst kind of
vulnerability that exists. By merely existing on the classpath, this
library causes the Java serialization parser for the entire JVM process
to go from being a state machine to a turing machine. A turing machine
with an exec() function!

https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2015-8103
https://commons.apache.org/proper/commons-collections/security-reports.html
http://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/

PR #822, opened 2017-04-10
Added LGPL v2.1 license file and section into README.md file

tfr42:
To prevent that we loose contributions due to missing license file added license section in README and in Maven default license.txt.
An example where a contribution was revoked is in https://github.com/deegree/deegree3/pull/687#issuecomment-195004790


tfr42:
The section "[Open Source License[(http://www.deegree.org/about-deegree#open-source-license)" updated accordingly. It missed the version for LGPL.

PR #823, opened 2017-04-13
CSW client: capability to send along custom http headers

awst-baum:
This enhancement adds the possibility to send custom HTTP headers along with the CSW client's GetRecords and GetRecordsById requests.


tfr42:
Thank you for your contribution. Unfortunately, it cannot be merged automatically. It cannot be considered for inclusion, before this has been fixed. Please rebase it against the target branch.

PR #832, opened 2017-05-19
startIndex fix for SQL featureStore (queryByOperatorFilter)

RomainWieser:
Pass the startIndex into SQL with OFFSET keyword and count whith LIMIT.

Not tested with other featureStore...


copierrj:
Thank you for your pull request.  Unfortunately your suggested improvement has a major drawback: it only works correctly for PostgreSQL. 

In order to support other databases, the deegree code contains different implementations of a concept called SQLDialect (see module: deegree-core-sqldialect). 

For this improvement to be accepted by the deegree project it is necessary to rework the code and have it use the sql dialect infrastructure instead of doing stuff directly in SQL. 

Feel free to ask if you have any questions.

Issue #827, opened 2017-05-09
DDL for PostgreSQL created with deegree contains too long column names 


lgoltz:
The DDL for HydroPhysicalWaters [1] created with the deegree-cli-utility [2] using deegree contains too long column names (e.g. gn_leastdetailedviewingresolution_gmd_md_resolution_gmd_eq_3137881). 

Expected is that the column names are truncated to the length specified with _maxLength_.

[1] http://inspire.ec.europa.eu/schemas/hy-p/4.0/HydroPhysicalWaters.xsd
[2] https://github.com/lat-lon/deegree-cli-utility/tree/deegree-3.4

dstenger:
Pull request: https://github.com/deegree/deegree3/pull/838

PR #838, opened 2017-06-02
Fix MappingContextManager to evaluate maxLength correctly

dstenger:
Previously, the maxLength set in MappingContextManager was not evaluated correctly leading to column names exceeding the length set by this parameter.
Issue was reported in https://github.com/deegree/deegree3/issues/827.

This fix enables that the column names are always truncated to the length specified by maxLength.


tfr42:
Thank you for your contribution. Your pull requests has been accepted by the TMC. Please find more information in the meeting chat available at https://github.com/deegree/deegree3/wiki/tmc-meeting-minutes-20171006.
This pull request will be part of the 3.4-RC4 release version.

PR #840, opened 2017-06-08
 Enhanced CRS configuration by CRS identifier in OGC HTTP URI syntax

### LINKED ISSUE #711: Support OGC HTTP URI style CRS references
deegree currently doesn't support CRS references like the following:

```
srsName="http://www.opengis.net/def/crs/EPSG/0/25832"
```

This notation resolves to the actual definitions provided by the OGC. It has been around for some time and is supported in other solutions like [GeoServer](http://docs.geoserver.org/stable/en/user/services/wfs/webadmin.html#gml).

The closest thing I could find as kind of official reference for this is an OGC Whitepaper from 2010 ([OGC 10-124r1](http://portal.opengeospatial.org/files/?artifact_id=39467)). Related to that is also this statement:

> In June 2010 OGC revised the naming policy to use http URIs to identify persistent OGC resources instead of URNs. The change to URIs was primarily to make resolution easier, and in order that OGC technologies can integrate with linked data and semantic web. 

Recent OGC documents (for instance the GML 3.3 spec) also feature this style of reference in all their GML examples.

I also heard from different sources that this is the notation style that should be used in INSPIRE GML documents, but haven't found any official reference for that yet.

lgoltz:
The CRS configuration was enhanced by the  CRS identifier in OGC HTTP URI syntax like `http://www.opengis.net/def/crs/EPSG/0/25832`. 

Fixes #711


tfr42:
Thank you for your contribution. Your pull requests has been accepted by the TMC. Please find more information in the meeting chat available at https://github.com/deegree/deegree3/wiki/tmc-meeting-minutes-20171103.
This pull request will be part of the 3.4-RC5 release version.

PR #850, opened 2017-10-02
Fix #848 [SQLFeatureStore] Deleting features fails for certain feature type prefix configurations

### LINKED ISSUE #848: [SQLFeatureStore] Deleting features fails for certain feature type prefix configurations
We observed that a delete by feature id against an SQLFeatureStore fails in some cases. This issue seems to be related to certain feature prefix configuration and differs between Java versions, e.g. we have two feature types with the following feature id prefices:

* APP_FEATURE_
* APP_FEATURE_X

We found that deleting the feature with id "APP_FEATURE_X_1" works in a Java 7 setup (tested with OpenJDK 7, 1.7.0_151 on Ubuntu 14.04 AMD64), while it fails with an exception on Java 8 (same machine, Oracle JDK 8, 1.8.0_144).

A first analysis points to method analyze(String) in class org.deegree.feature.persistence.sql.id.IdAnalyzer:

<pre>
...
        // TODO implement this more efficiently
        for ( String prefix : prefixToFt.keySet() ) {
            if ( featureOrGeomId.startsWith( prefix ) ) {
                ft = prefixToFt.get( prefix );
                break;
            }
        }
...
</pre>

We believe this is caused by the different key order in the respective HashMap implementations:

https://stackoverflow.com/questions/38587711/hashmap-keyset-returns-keys-in-different-order-in-java-7-java-8

nitram509:
Hi,

this PR addresses issue #848 [SQLFeatureStore] Deleting features fails for certain feature type prefix configurations.
When one searches features by a prefix, the search operation has to be done on a "stable" == "sorted" input list. The issue already mentioned, the undefined order behavior of hash-map keys leads to this issue.
The included test fill fail, if one removes the sorting method. (tested on Java 1.6.0_45).

I did some additional code cleanup in second commit, as well.

Any feedback is welcome.

Keep rocking this cool project 👍.
Thank you in advance.


MrSnyder:
That's funny, we seem to have been working on the exact same issue. :-)

On first glimpse, our approaches seems to be very similar and I believe they should be behave effectively the same (see #854 and #855). What is your impression?

Have a nice weekend,
Markus

nitram509:
That's funny indeed :-)

I see, your implementation uses the same approach to solve the problem 😀 
Frankly said, about our different implementations I have no preference, they seem to me equivalent.

The only thing which is not clear to me, is that IMHO your test doesn't fail on wrong/previous implementation (honestly not tested, just from looking at the code).
I propose to change the setup in a way, you create one colliding situation.
It took me a while to find one and you can try my test-setup - where 'doubleLetterTest' fails, when implementation is wrong.

And good catch about the 'no match' situation  👍, I've overseen

Makes sense?

Have a nice weekend as well,
Martin

PS: I would be proud & glad if my PR is accepted, but I'm also fine with your solution ;-)

MrSnyder:
Martin,

thanks for your quick reply. I finally found the time to take a thorough look at your code and I would like to suggest to merge our efforts. Therefore, I rebuilt my bugfix branch based on your PR branch [1] and applied minor modifications, mostly to make my customer happy (who needs a specific error in production to go away):

- In IdAnalyzer, sorting of keys is only done once in the constructor now.
- I based IdAnalyzerTest on my version of this class, because I found it a bit more thorough already and it matches with my extended description in ticket #848. Btw, the test for the colliding situation is in analyzeFeatureIdMultiplePrefixMatches (I added a comment to point that out). Also, my customer would like to see the actual production issue tested, so I added it to this method as well.

If you like this route, please pull my branch [1] so the PR gets updated. As soon as we're both happy with the PR, I will derive a backport for deegree 3.3. 

[1] https://github.com/occamlabs/deegree3/tree/bugfix/sqlfeaturestore/idanalysis-ambiguity

nitram509:
Looks good to me.
Thank you for that co-work 👍

nitram509:
PS: sorry, accidentally closed.

MrSnyder:
Hi Martin,

there seems to be an issue on Travis CI due to their change in the standard linux distribution. I tried to fix the .travis.yml, can you please pull my branch again?

Best regards,
Markus

MrSnyder:
Hi Martin,

please pull my branch again. I am trying to solve the Travis CI build issue on OpenJDK 7 -- seems to be caused by a change in the linux distribution Travis CI uses by default.

Thanks!

nitram509:
Done.
🖖

tfr42:
Please rebase to current master. Travis-ci issue is fixed, see #852.

nitram509:
Rebased to master, as proposed.

MrSnyder:
Hi Martin,

sorry for the confusion, but could you remove the "dist: precise"  line from the .travis.yml again? This was just my shot at working around the build issue and probably doesn't make sense after Torsten's changes in master.

Best regards,
Markus

nitram509:
Done.
🖖

MrSnyder:
Hi Martin,

thanks a lot. Good to see that the CI checks work again.

One last thing: The PR currently includes 50 commits and 152 file changes...this doesn't seem right. It probably happened when rebasing the PR to master. Could you clean this up? If you need help, please let me know... I would probably create a new branch and cherry-pick the relevant commits, but there may be a more elegant way to do it.

Best regards,
Markus

nitram509:
I'm closing this PR and create successor #867 instead. This new PR contains just the relevant commits and changed files.
https://github.com/deegree/deegree3/pull/867

PR #854, opened 2017-10-06
Bugfix for #848: Ensure unambigous results when determining feature type from id

MrSnyder:
This is a bugfix for #848:

* Ensures that the feature id match is always based on the longest matching prefix
* Includes tests


MrSnyder:
Withdrawn in favor of #850.

PR #867, opened 2017-11-15
Fix #848 - new [SQLFeatureStore] Deleting features fails for certain feature type prefix configurations

### LINKED ISSUE #848: [SQLFeatureStore] Deleting features fails for certain feature type prefix configurations

nitram509:
Hi,

this PR supersedes #850.
It reduces the PR to the actual changed files and relevant commits.

Fore actual content details, please have a look at https://github.com/deegree/deegree3/pull/850

Regards
Martin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,3 @@ env: MAVEN_OPTS="-XX:MaxPermSize=256m"
 cache:
   directories:
     - $HOME/.m2/repository/
-dist: precise
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,5 @@ env: MAVEN_OPTS="-XX:MaxPermSize=256m"
 cache:
   directories:
     - $HOME/.m2/repository/
+dist: precise
+
--- a/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/utils/ArrayUtils.java
+++ b/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/utils/ArrayUtils.java
@@ -38,11 +38,7 @@ package org.deegree.commons.utils;
 import static java.lang.Double.parseDouble;
 import static java.lang.Float.parseFloat;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.StringTokenizer;
+import java.util.*;
 
 /**
  * This is a collection of some methods that work with arrays and lists, like join or removeAll. It is complementary to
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -12,4 +12,4 @@ script: "travis_retry mvn verify -Pintegration-tests -B -V"
 env: MAVEN_OPTS="-XX:MaxPermSize=256m"
 cache:
   directories:
-    - $HOME/.m2/repository/
\ No newline at end of file
+    - $HOME/.m2/repository/
```
tfr42: Can you please remove this change!

copierrj:
Thank you for your contribution. The TMC meeting noticed that your PR contains a change to the .travis.yml file. It was therefore decided that we would to ask you if you can remove this change.

nitram509:
@tfr42 sure, I'm willing to support and follow your advice.
BUT: I hesitate to do it, because I believe there is a miss-understanding.

This change was introduced 9/Feb/2015 and I did not touched/changed this line. I did merge from MASTER and there you can see, this change is in.
I believe there is a visualization issue in the Github Web-UI, which wrongly shows that I have added this line - which is not the case.

Indeed, I removed a line from .travis.ml
>>> dist: precise
(according to comment in #850).

Would you please double check and confirm,
that I should remove the caching configuration for the Maven repo-folder?

tfr42:
> Would you please double check and confirm,

Looks ok. It seems to be a CR/LF character issue.

> that I should remove the caching configuration for the Maven repo-folder?

No, please leave it as it is.

PR #914, opened 2018-07-20
Update pom.xml

copierrj:
Upgrade deegree maven plugin to version 2.0.2


tfr42:
Root cause for the failing build seems to be the version range defined for the transitive dependency of net.postgis:postgis-jdbc:2.2.1, see https://repo1.maven.org/maven2/net/postgis/postgis-jdbc/2.2.1/postgis-jdbc-2.2.1.pom
Issue should be fixed with PR #932.

copierrj:
Looks like we've still another build issue...

PR #921, opened 2018-09-07
Upgrades Tomcat bundle to 8.5.x line since 8.0.x line has reached EOL.

tfr42:
This PR upgrades the bundle of Apache Tomcat to 8.5 line. 8.0 line has reached end-of-live.


tfr42:
It requires update to Tomcat 8.5.35 due to http://tomcat.apache.org/security-8.html#Fixed_in_Apache_Tomcat_8.5.32

PR #949, opened 2019-01-17
Upgrade JTS from version 1.13 to 1.16.0

tfr42:
This PR migrates JTS from version 1.13 to 1.16.0 following the migration guidelines from JTS as described in https://github.com/locationtech/jts/blob/master/MIGRATION.md.
Furthermore the PR fixes the missing definition for compile dependencies in some sub-modules.

### DIFF
--- a/deegree-core/deegree-core-filterfunctions/pom.xml
+++ b/deegree-core/deegree-core-filterfunctions/pom.xml
@@ -34,6 +34,10 @@
       <artifactId>deegree-core-base</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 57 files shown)


tfr42:
Requires an integration test with a database as a backend:
- PostgreSQL 9.6 with PostGIS 2.5

Test scenario
- WFS-T INSERT
- WFS GETFEATURE
- WMS GETMAP with different CRS
- on DB level `st_asewkt(geometry)`

tfr42:
Tools to use for setting up an integration test may be soap ui which supports HTTP requests of any kind and SQL statements as well.

Issue #942, opened 2018-12-03
Security issue: Local file inclusion in deegree3 console


olis84:
deegree3 console allows via ./console/client/download.jsp to an unauthorized user to download files from server. The file name isn't validated sufficiently, so hackers are able to get local files. It's also possible to download the password from the interface of deegree3 console. 
It would be recommended to validate the user's input. Values like '../' should be filtered to prevent access to higher-level folders. During a penetration test, it was possible to access /etc/passwd an other sensible files on host.

tfr42:
Thanks for your investigation. Please have a look on our [guidelines](https://github.com/deegree/deegree3/wiki/Bug-Report-writing-Guidelines#reporting-a-security-vulnerability) for future posting.

tfr42:
There is a SNAPSHOT version available including a fix ready for testing:
- WAR file: http://buildserver.deegree.org/view/misc/job/deegree-3.4-PR955/ws/deegree-services/deegree-webservices/target/deegree-webservices-3.4.4-SNAPSHOT.war
    - download and then just deploy it manually to your preferred web container
- deegree bundled with Tomcat: 
http://buildserver.deegree.org/view/misc/job/deegree-3.4-PR955/ws/deegree-services/deegree-webservices-tomcat-bundle/target/deegree-webservices-tomcat-bundle-3.4.4-SNAPSHOT-distribution.zip
    - download and then run the following commands:
```
unzip deegree-webservices-tomcat-bundle-3.4.4-SNAPSHOT-distribution.zip 
cd deegree-webservices-tomcat-bundle-3.4.4-SNAPSHOT/apache-tomcat-8.0.50/bin/
chmod a+x *.sh
./startup.sh
```
- Docker Image: https://cloud.docker.com/repository/docker/tfr42/deegree/general
```
docker pull tfr42/deegree:3.4.4-SNAPSHOT
docker run -p 8080:8080 --name deegree --rm tfr42/deegree:3.4.4-SNAPSHOT
```

tfr42:
Official version 3.4.4 available including the bug fix. Download from http://www.deegree.org/Download

PR #958, opened 2019-02-17
Updating handbook with security how to

### LINKED ISSUE #943: Security issue: Missing input validation for upload via console
Uploading new ZIP-workspaces via the console (./config/upload) is a helpful feature, but there is no input validation of the upload path. The ZIP file will be unzipped after upload directly to the upload path. 
It would be recommended to validate the user's input. Values like '../' should be filtered to prevent access to higher-level folders. During a penetration test, it was possible to upload files to the /tmp folder.
See also https://www.owasp.org/index.php/File_System#Path_traversal

tfr42:
Fixes #943 and #942  Added  new section to handbook on how to run securely deegree in production environments.


tfr42:
Documentation is online: http://download.deegree.org/documentation/3.4.4/html/installation.html#securing-deegree

PR #972, opened 2019-04-25
Remove check of area of use of CRS during insert

### LINKED ISSUE #969: Check if geometries are inside the valid domain of a crs should be optional (WFS-T)
Geometries may overlapping the valid domain of multiple crs, e.g. EPSG;25832 and EPSG:25933 During insert an failure occurred: "At least one geometry is not in the valid domain of the srs."

It is expected that the check can be enabled by configuration (disabled per default).

dstenger:
This pull request removes the check of area of use of CRS during insert.

This solves #969 partly.

Pull request #971 adds a configuration option to enable the check of area of use of CRS during insert by configuration.


lgoltz:
This PR solves also #973.

PR #999, opened 2019-07-18
Added support of XSLT 2.0

dstenger:
This PR adds the support of XSLT 2.0 for the transformation to HTML of GetFeatureInfo responses as well as GetCapabilities and Exceptions.

Furthermore, the xalan dependency was removed and xerces updated.

### DIFF
--- a/deegree-core/deegree-core-base/pom.xml
+++ b/deegree-core/deegree-core-base/pom.xml
@@ -39,10 +39,6 @@
       <artifactId>commons-codec</artifactId>
     </dependency>
     <dependency>
-      <groupId>xerces</groupId>
-      <artifactId>xercesImpl</artifactId>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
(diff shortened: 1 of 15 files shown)


copierrj:
It looks like this PR is also changing the default XSLT version. This would result in uses having to update their existing XSLT documents by adding a version="1.0" attribute.

deegree-core/deegree-core-base/pom.xml
```diff
@@ -38,10 +38,6 @@
       <groupId>commons-codec</groupId>
       <artifactId>commons-codec</artifactId>
     </dependency>
-    <dependency>
```
copierrj: Why removing this specific xerces dependency?

lgoltz:
> It looks like this PR is also changing the default XSLT version. This would result in uses having to update their existing XSLT documents by adding a version="1.0" attribute. Breaking existing workspaces should be avoided.

The version attribute is required since 1.0 (see https://www.w3.org/1999/11/xslt10.dtd). Saxon is more strict then the the default JAXP implementation and fails if the xslt file is not valid. I think we shouldn't allow invalid xslt files (as before).

deegree-core/deegree-core-base/pom.xml
lgoltz: xerces is used only in deegree-core-commons.

tfr42:
You may need to set a environment variable (system property) as documented in #1110.

PR #1004, opened 2019-09-19
Enhanced implementation for WFS 2.0 response paging

### LINKED ISSUE #648: Enable handling of wfs parameters startindex and count in datastores
For database backed datastores it would be nice to implement handling of startindex and count by adding the appropriate stuff into the sql query.

tfr42:
Fixes #648 and replaces #832. This pull request enhances the paging mechanism for WFS GetFeature response processing when parameter STARTINDEX and COUNT are provided.
It uses SQL `OFFSET/ROWS FETCH NEXT` syntax which requires at least Oracle 12c, MS SQLServer 2008 or PostgreSQL 9.4. The behavior of PR  #619 is preserved.


copierrj:
Unfortunately we had to revert this pull request because it was discovered that it breaks the build. In order to accept the suggested change it is required to prepare a new pull request. Sorry about that.

PR #1001, opened 2019-09-06
Migrate handbook from Sphinx to asciidoc

lgoltz:
Fixes #858

PR #1005, opened 2019-09-20
Enable configuration of multiple schema locations

### LINKED ISSUE #892: Return original GML 3.1.1 and 3.2.1 schemas for a single FeatureStore and WFS endpoint
Our aim is to return the description of FeatureTypes of a single FeatureStore (operation: DescribeFeatureType) in different GML formats.
The GML schemas should not be generated by deegree but original, provided schemas should be used.
What is the best approach?

dstenger:
This enhancement makes it possible to configure multiple schema locations to support static schemas for multiple GML versions in DescribeFeatureType responses of a single WFS.

For example, if DescribeFeatureType responses of a single WFS shall support non generated schemas for multiple GML versions, the newly introduced configuration option solves that use case.

Please also see #892 for further details.

Caution, this pull request also includes #1001. Thus, #1001 should be merged first.

Fixes #892.


tfr42:
The prerequisite that PR #1001 is merged is fullfilled.

Issue #886, opened 2018-02-21
GeometryValidator: incorrect result for polygon in left-handed CRS


jechterhoff:
Hello,

The GeometryValidator produces incorrect results when checking the orientation of the exterior and interior boundaries of a surface whose coordinates are given in a left-handed CRS. I've added a small test program at the end of this message.

Apparently, the determination of whether a given ring is counter-clockwise or not using the JTS CGAlgorithms.isCCW(...) in GeometryValidator.java, method validatePatch(...), requires that the coordinates are given in a right-handed CRS. 

The OGC CITE tests transform the coordinates of a ring to a right-handed CRS, if necessary. See https://github.com/opengeospatial/ets-gml32/blob/master/src/main/java/org/opengis/cite/iso19136/data/spatial/GeometryAssert.java and https://github.com/opengeospatial/geomatics-geotk/blob/master/src/main/java/org/opengis/cite/geomatics/GeodesyUtils.java (search for transformRingToRightHandedCS).

The big question is: How could/should the GeometryValidator be fixed? Or am I missing something?

I've tried a similar approach as CITE, relying solely on Apache SIS. That works, but it would introduce a dependency that may not be necessary. Can deegree realise the necessary transformation on its own?

If the answer is no, could we introduce a template method in GeometryValidator.java, which would be called before invoking the orientation check? Like:

```java
Coordinate[] transCoords = transformRingToRightHandedCS(ring_coords,crs);
boolean isClockwise = !CGAlgorithms.isCCW(transCoords);
...
public Coordinate[] transformRingToRightHandedCS(Coordinate[] coords, ICRS crs) {+
// Default is to simply return the coordinates.
return coords;
}
```
Then a subclass of GeometryValidator could override the method and transform the coordiantes to a right-handed CRS, using - for example - Apache SIS.

Comments and suggestions are welcome.

Best regards,
Johannes

```java
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;

import org.deegree.cs.coordinatesystems.ICRS;
import org.deegree.cs.persistence.CRSManager;
import org.deegree.geometry.Geometry;
import org.deegree.geometry.GeometryFactory;
import org.deegree.geometry.primitive.Ring;
import org.deegree.geometry.standard.points.PackedPoints;
import org.deegree.geometry.validation.GeometryValidationEventHandler;
import org.deegree.geometry.validation.GeometryValidator;
import org.deegree.geometry.validation.event.ExteriorRingOrientation;
[...]

Issue #1033, opened 2020-02-07
Test of deegree-webservices-3.4.12-SNAPSHOT


dstenger:
Current snapshot build shows regression errors 
- with Oracle JDK 1.8_212 - http://buildserver.deegree.org/job/deegree-3.4-master/org.deegree$deegree-wms-similarity-tests/208/testReport/junit/org.deegree.services.wms/WMSSimilarityIntegrationTest/testSimilarity_34_/ 
- same with OpenJDK 8 - http://buildserver.deegree.org/view/misc/job/deegree-3.4-openjdk8/lastBuild/org.deegree$deegree-wms-similarity-tests/testReport/org.deegree.services.wms/WMSSimilarityIntegrationTest/testSimilarity_34_/ and 
- and Corretto 8 - http://buildserver.deegree.org/view/misc/job/deegree-3.4-Corretto8/lastBuild/org.deegree$deegree-wms-similarity-tests/testReport/org.deegree.services.wms/WMSSimilarityIntegrationTest/testSimilarity_34_/ 

Issue was discussed during TMC meeting https://github.com/deegree/deegree3/wiki/tmc-meeting-minutes-20200207

dstenger:
# Setup

* TEAM Engine hosted by OGC is used for testing.
* URL: https://cite.opengeospatial.org/te2/
* Username: deegree

# Results

## WFS 2.0

* Tested endpoint: http://cite.deegree.org/deegree-webservices-3.4.12-20200124.110857-4/services/wfs200?service=WFS&request=GetCapabilities
* All mandatory tests pass.
* Result: Fail 9 / Pass 87 / Skip 32 / Total 128
* Test run: https://cite.opengeospatial.org/te2/viewSessionLog.jsp?session=s0001
* Comparison with deegree reference implementation:
  * http://cite.deegree.org/deegree-webservices-3.4.5/services/wfs200?service=WFS&request=GetCapabilities
  * Result: Fail 1 / Pass 95 / Skip 32 / Total 128
  * https://cite.opengeospatial.org/te2/viewSessionLog.jsp?session=s0002

Test of current SNAPSHOT passes less tests.
Following tests fail in addition:
* after Instant (2x): No temporal values found in results: property is "urn:x-inspire:specification:gmlas:ProtectedSites:3.0":legalFoundationDate.
* after Period (2x): No temporal values found in results: property is "urn:x-inspire:specification:gmlas:ProtectedSites:3.0":legalFoundationDate.
[...]

PR #1034, opened 2020-02-10
Fixes issue #886

carlospzurita:
Added fix for https://github.com/deegree/deegree3/issues/886 methods `hasLeftHandedSrs`, `isInterior` and `isExterior` in InteriorRingOriention and ExteriorRingOrientation respectively. The solution implemented checks the orientation of the axis for the declared Polygon patch to decide the left-handness, and uses this to check if the ring is exterior or interior.

### DIFF
--- a/deegree-core/deegree-core-base/src/test/java/org/deegree/geometry/validation/GeometryValidatorTest.java
+++ b/deegree-core/deegree-core-base/src/test/java/org/deegree/geometry/validation/GeometryValidatorTest.java
@@ -49,7 +49,9 @@ import javax.xml.stream.XMLStreamException;
 import junit.framework.Assert;
 
 import org.deegree.commons.xml.XMLParsingException;
+import org.deegree.cs.coordinatesystems.ICRS;
 import org.deegree.cs.exceptions.UnknownCRSException;
+import org.deegree.cs.persistence.CRSManager;
 import org.deegree.geometry.Geometry;
 import org.deegree.geometry.validation.event.CurveDiscontinuity;
 import org.deegree.geometry.validation.event.CurveSelfIntersection;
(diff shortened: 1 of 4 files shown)


carlospzurita:
We are seeing some errors related to the build process in Travis and Jenkins. We have looking into the logs and have seen some errors that also appeared in other PR. Is there any issue with the continuous integration pipeline?

copierrj:
Travis is currently broken (= fails even when local builds and Jenkins do succeed). This happens often and we are currently unsure what to do about this.

Unfortunately there is also a known regression (#1033) in master. We are currently working on that.

copierrj:
Regarding both hasLeftHandedSrs() methods: these two are almost identical. Please create an appropriate helper method in AbstractGeometryValidationEvent in order to avoid too much code duplication.

carlospzurita:
We added the modifications indicated, both in the AbstractGeometryValidationEvent and in the GeometryValidatorTest.

The test for validation is not fully complete, given that in order to maintain the compatibility with other clients, we did not include the left-handed and right-handed in the validation process for geometry orientation.

carlospzurita:
We see that there are some errors in the CI processes, but we are not sure if this is related to our own code uploaded, the tests provided or other unrelated issue. Can you provide us with some feedback about this? Thank you

carlospzurita:
Would it be useful for the integration of the pull request if we attend the TMC meeting on 20/03/2020? We can provide any feedback that you need for the solution implemented or the tests included.

copierrj:
We've looked at the latest changes you made and we are happy with the result.

MarcoMinghini:
@tfr42 do you know when the next deegree version including this fix will be released?

PR #1042, opened 2020-02-27
Introduce CSV as output format for WFS

dstenger:
This pull request introduces CSV as output format for WFS.


tfr42 (changes requested):
License header are missing

PR #1048, opened 2020-03-25
enhance similarity test for easier problem solving

stephanr:
* output failed images to log/stdout as base64 encoded zip of the failed image
   * optimize test for debugging in developer ide
   * prevent adding of question mark if it is already part of the request
   * optimize log output


stephanr:
I have added a commit to reenable testing of Java2DRendererTest to make the previously disabled (manual) tests available.

stephanr:
Error Images
textstyling2
textstyling

PR #1055, opened 2020-04-06
Skip GML reference resolving for WFS-T

dstenger:
This PR adds a new configuration option `UseExistingSkipResolvingReferences` (configuration example: `<EnableTransactions idGen="UseExistingSkipResolvingReferences" >true</EnableTransactions>`) making it possible that references to features are not checked when inserting a new feature via WFS-T.


tfr42 (changes requested):
License header is missing

PR #1067, opened 2020-06-04
Enable support of transparency for a 4-band GeoTIFF used with GdalLayer

### LINKED ISSUE #1065: Using 4-band GeoTIFF with GdalLayer is not supporting transparency
Using a 4-band (RGBA) GeoTIFF with GdalLayer is not supporting the alpha values. Instead background color is shown.

dstenger:
This pull request enables support of transparency for a 4-band GeoTIFF used with GdalLayer.

Detailed description of problem: #1065

Closes #1065.


copierrj:
I'm not sure if we really want to just drop this transparency band stripping code to fix this issue. It was (most likely) added to solve another issue (probably related to handling requests with JPEG as output format). Are you sure we are not breaking something?

stephanr:
From my point of view, with this change we run the risk to help those who want to use a raster with RGB + Alpha but for users who only use the RGB part of RGB + Infrared this would cause problems.

Perhaps it would be better to make this configurable or look up what was requested.

PR #1068, opened 2020-06-16
Export original XML schema if exactly one local schema is configured

### LINKED ISSUE #1044: DescribeFeatureType reponse of deegree WFS causes ArcGIS online to fail
The issues we've been having is that most of our services using deegree either do not open, or do not show any data when using ArcGIS online as a client. We don't get any useful error messages back unfortunately.
We think it's a DescribeFeatureType issue, so we've been doing some more research (with both QGIS and ArcGis Online) and what we've found out is that the configuration of `DisableDynamicSchema` is relevant. Either the returned schema is not valid or the clients can not process the response of the DescribeFeatureType request.
We're assuming is that the only way our services can work in ArcGIS online (and with a schema that validates), is if the response from a DescribeFeatureType request is a full schema coming directly from the file system or external resources.

dstenger:
If only one schema is configured in a FeatureStore, the schema is referenced in the local file system and DisableDynamicSchema is set to true, a DescribeFeatureType request returns the original XML schema instead of including the schema in a deegree generated schema via include.

Closes #1044.

### DIFF
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/schema/GMLAppSchemaWriter.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/schema/GMLAppSchemaWriter.java
@@ -35,6 +35,59 @@
  ----------------------------------------------------------------------------*/
 package org.deegree.gml.schema;
 
+import org.apache.xerces.xs.StringList;
+import org.apache.xerces.xs.XSAttributeUse;
+import org.apache.xerces.xs.XSComplexTypeDefinition;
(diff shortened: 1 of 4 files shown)


stephanr:
As a TMC, we would prefer the sorting of the Java import statements to remain the same.
We would like to keep the formatting to prevent future problems such as merge conflicts.

copierrj:
The TMC agrees that limiting our use of more advanced XML schema features to deal with limited XML schema implementations in clients is a good thing. Therefore we endorse the proposed change.

PR #1084, opened 2020-09-16
Fix cache handling of raster files

stephanr:
This PR contains both a correction of the cache handling and a method to better control the raster cache for file base raster.

Previously there was the problem that the naming of cache files was derived from the raster data.
This could lead to unwanted errors in two situations.  
First it can happen that raster files in different directories have identical names (For example ../data/dop/overview.tif and ../data/infrared/overview.tif).  
Secodn the cache files were not distinguished between the different levels, which e.g. a geotiff can contain, and the same cache file was rewritten on every access.

### DIFF
--- a/deegree-core/deegree-core-coverage/src/main/java/org/deegree/coverage/raster/cache/RasterCache.java
+++ b/deegree-core/deegree-core-coverage/src/main/java/org/deegree/coverage/raster/cache/RasterCache.java
@@ -486,7 +486,7 @@ public class RasterCache {
                 boolean createCache = reader.shouldCreateCacheFile();
                 File cacheFile = null;
                 if ( createCache ) {
-                    LOG.trace( "create cachefule for location {}", reader.getDataLocationId() );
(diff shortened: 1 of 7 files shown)


deegree-core/deegree-core-coverage/src/main/java/org/deegree/coverage/raster/cache/RasterCache.java
```diff
@@ -481,6 +486,7 @@ public RasterReader addReader( RasterReader reader ) {
                 boolean createCache = reader.shouldCreateCacheFile();
                 File cacheFile = null;
                 if ( createCache ) {
+                    LOG.trace( "create cachefule for location {}", reader.getDataLocationId() );
```
tfr42, line 489: minor typo in "cachefule" to "cachefile"

stephanr: fixed in https://github.com/deegree/deegree3/pull/1084/commits/61cc5eb3ca48965babd2eac22196db3d6ebf30db

PR #1087, opened 2020-09-16
Allow rewriting of request parameters for KVP request

stephanr:
This PR is a supplement to #1086 to manipulate single KVP parameters.

The reason for this is, that some desktop clients are sending invlaid parameters or expection uncommon default values.  
With this PR parameters of a KVP request can be manipulated to replace faulty parameters or to enable selective defaulting. 

For example invalid `DPI` parameters can be removed or a default value can be set (as `X-DPI`).

This is targeted to be an easier replacement for rewriting request with a proxy-server like nginx or apache.


copierrj:
Improving the DPI handling situation in deegree is appreciated. However, the TMC is not ready to accept functionality only to deal to misbehaving clients.

stephanr:
> Improving the DPI handling situation in deegree is appreciated. However, the TMC is not ready to accept functionality only to deal with misbehaving clients.

Since the DPI parameter handling has already been merged with #1086, the rewrite functionality is not continued and the PR will be closed.

Issue #1096, opened 2020-10-29
ConcurrentModificationException while destroying single resources


tfr42:
Currently the integration tests fail in module "deegree-resource-deps-tests" due to the following errors:
```
16:15:38]  INFO: [CRSManager] Clear CRS store and transformation map
[16:15:38]  INFO: [CRSManager] --------------------------------------------------------------------------------
[16:15:38]  INFO: [CRSManager] No 'crs' directory -- use default configuration.
[16:15:38]  INFO: [CRSManager] --------------------------------------------------------------------------------
[16:15:38]  INFO: [CRSManager] Setting up crs store 'default' from file 'file:/var/jenkins_home/workspace/deegree-build-upload/deegree-core/deegree-core-cs/target/deegree-core-cs-3.4.14-SNAPSHOT.jar!/org/deegree/cs/persistence/default.xml'...
[16:15:38]  INFO: [CRSManager] Registering global crs store with id 'default', type: 'org.deegree.cs.persistence.deegree.d3.DeegreeCRSStore'
[ERROR] Tests run: 8, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 8.576 s <<< FAILURE! - in org.deegree.workspace.SimpleSqlFeatureStoreTest
[ERROR] testDestroyInitializeSingle(org.deegree.workspace.SimpleSqlFeatureStoreTest)  Time elapsed: 0.966 s  <<< ERROR!
java.util.ConcurrentModificationException
	at org.deegree.workspace.SimpleSqlFeatureStoreTest.testDestroyInitializeSingle(SimpleSqlFeatureStoreTest.java:156)

[ERROR] testDestroySingle(org.deegree.workspace.SimpleSqlFeatureStoreTest)  Time elapsed: 0.976 s  <<< ERROR!
java.util.ConcurrentModificationException
	at org.deegree.workspace.SimpleSqlFeatureStoreTest.testDestroySingle(SimpleSqlFeatureStoreTest.java:144)

[ERROR] testReinitializeChain(org.deegree.workspace.SimpleSqlFeatureStoreTest)  Time elapsed: 0.665 s  <<< ERROR!
java.util.ConcurrentModificationException
	at org.deegree.workspace.SimpleSqlFeatureStoreTest.testReinitializeChain(SimpleSqlFeatureStoreTest.java:170)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[...]

PR #1098, opened 2020-11-13
#1096 set failing tests to ignore, this is a temporary solution

tfr42:


tfr42:
Decided during todays TMC meeting to overcome the current blocking situation.

tfr42:
fixes #1096 by ignoring the affected unit tests

PR #1111, opened 2020-12-21
Enhance AppSchemaMapper by reference data

dstenger:
This pull requests adds a new option to the deegree-cli-utility tool to consider reference data.

New functionality is described here: https://github.com/lat-lon/deegree3/blob/configCliWithReferenceData-6535/deegree-services/deegree-webservices-handbook/src/main/asciidoc/cli-utility.adoc#usage-of-option-referencedata


lgoltz:
Will be included in the new GmlTool (s. #1222).

PR #1118, opened 2021-01-19
fix count limit for response paging in two situations

stempler:
1. Fix behavior when count provided for response paging is higher than the maximum features limit that is configured.
Previously when using a count that is higher than the maximum features that can be retrieved, the request URIs generated for the next requests would have gaps, as the start index would still be increased by the given count and the count would be retained as well.

2. Limit the count for the previous paged response in case the difference to the current start index is less than the original count.


tfr42:
Thank you for your contribution. Unfortunately, it cannot be merged automatically. It cannot be considered for inclusion, before this has been fixed. Please rebase it against the target `main` branch.

PR #1128, opened 2021-02-25
Enhanced implementation for WFS 2.0 response paging

lgoltz:
Renews reverted pull request #1004


tfr42:
This PR is included in #1206

Issue #773, opened 2016-09-19
Replace deegree-compliance-tests by official and final OGC ETS


tfr42:
Currently the OGC CITE Tests in deegree-tests/deegree-compliance-tests for
- WCS 1.0.0
- WFS 1.0.0, 1.1.0
- WMS 1.1.1, 1.3.0
are based on quite old forks of the related OGC ETS. Prefered setup is based on most recent release version of the official OGC CITE ETS.
Including also ETS for:
- WFS 2.0
- WMTS 1.0
- CSW 2.0.2
- WPS 1.0

tfr42:
CLI available in OGC TEAM engine: https://github.com/opengeospatial/cite/wiki/Conformance-Testing-with-TestNG-Part-2#12-2-command-shell-console

tfr42:
The OSGeo Foundation has started an OGC Certification Service to provide a public demonstration instance used for the OGC TEAM Engine to be executed against. See the wiki https://wiki.osgeo.org/wiki/OGC_Certification_Services for more information.

dstenger:
Disadvantage when decoupling the CITE integration tests from Maven build: Tests cannot be executed against feature branches used for pull requests.

Possible workflows (decoupled from Maven build):

* Workflow 1: Usage of Docker (independent from OGC infrastructure).
  * 1. Create and start Docker Container with latest deegree artifact.
    * deegree CITE workspace is integrated into the container.
  * 2. Create and start Docker Container of OGC test suites.
    * These are available via Docker Hub (https://hub.docker.com/u/ogccite).
  * 3. Execute SoapUI tests which use the REST interface of the OGC test suites to test the deegree endpoints.
    * SoapUI parses the responses and applies assertions.
    * SoapUI tests can be started headless either via Maven or CLI.
    * Alternatively, Java can be used instead of SoapUI.

* Workflow 2: No usage of Docker (depends on OGC infrastructure).
  * 1. Start the latest deegree artifact with Java (e.g. via using Maven plus plugin).
    * deegree CITE workspace can be accessed by deegree instance.
  * 2. Execute SoapUI tests which use the REST interface of the OGC infrastructure (https://cite.opengeospatial.org/teamengine/) to test the deegree endpoints.
    * See sub bullet points of step 3 of workflow 1.

* Workflow 3: No usage of Docker (independent from OGC infrastructure).
  * 1. Start the latest deegree artifact with Java (e.g. via using Maven plus plugin).
[...]

PR #1153, opened 2021-06-28
Added soapUI test module deegree-integrations-tests

lgoltz:
This PR adds a new module called "deegree-acceptance-tests" which contains a soapUI (www.soapui.org) maven plugin configuration executing a soapUI test script against a local instance of deegree webservices. The test script verifies the REST-API, and deegree webservices endpoint operations.
This test script is intended to contain tests for deegree specifc extensions and resources. It is complementary to the proposed replacement of the current built-in (but outdated) integration tests (see issue #773).


copierrj:
The TMC likes the approach of this PR. However, we would prefer the new module to be named differently. We do not consider these tests to be acceptance. Please change the name of the module into something else such as deegree-soapui-tests. 

Because test like these are hard to review for the TMC we would like a list of what is actually tested included. Could you add a small readme document for this purpose to the module?

tfr42 (changes requested):
change to the module name

PR #1162, opened 2021-07-29
Enhance GeoJson writer

dstenger:
Enhance GeoJSON output by lists of primitive and complex properties.

Caution: This pull request bases on https://github.com/deegree/deegree3/pull/1161 which should be merged first.

PR #1164, opened 2021-07-29
Implementation of GeoJsonSingleFeatureWriter

dstenger:
GeoJsonSingleFeatureWriter is needed for the implementation of OGC API Features.

Caution: This pull request bases on https://github.com/deegree/deegree3/pull/1162 which should be merged first.


stephanr:
During the review of the PR, it was noticed that the copyright is not present on all source files. Accordingly, we would appreciate adding these later.

See also issue https://github.com/deegree/deegree3/issues/1348

PR #1167, opened 2021-08-10
Refactored import statements to use org.junit.Assert 

tfr42:
This PR contains minor improvements of the usage of JUnit API.

- [x] refactored import statements to use org.junit.Assert instead of deprecated junit.framework.Assert type
- [x] uses assertEquals for double values with a delta of `0.01d`, see https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertArrayEquals(double[],%20double[],%20double)
- [x] upgraded junit to 4.13.2 (see related PR #1089)


tfr42:
Introduce constants for different units and measures.

stephanr:
From a programmer's point of view it would be nice to use general available constant for most cases and only a local constant if required.
It might also be interesting to derive the general constant from the value previously used by JUnit.

Issue #1079, opened 2020-08-14
Remove travis build job (reproducer) and replace badge with Jenkins build status


tfr42:
- [x] Remove http://buildserver.deegree.org/view/Failing%20Jobs/job/internal/job/travisci-build/
- [x] remove travis-ci pipeline from project
- [x] Install embeddable-build-status-plugin for Jenkins
- [x] Replace the badge for current build status in README.md
- [ ] ~~Remove https://github.com/deegree/deegree3/tree/master/.github/workflows~~
- [ ] ~~disable GitHub workflows~~

tfr42:
Job is deactivated http://buildserver.deegree.org/view/Failing%20Jobs/job/internal/job/travisci-build/

tfr42:
Build status plugin installed for Jenkins http://buildserver.deegree.org/job/deegree-3.4-master/badge/

tfr42:
Changes to README.md in https://github.com/deegree/deegree3/pull/1082

tfr42:
PR #1179  is merged

tfr42:
The new server infrastructure  #1130 is up and running, including recent Jenkins version building all PR https://buildserver.deegree.org/view/misc/job/deegree-3.4/job/pull-request/view/change-requests/

tfr42:
The GitHub action was re-activated to have a second "opinion" when building PR. GitHub action runs the maven "test" lifecycle phase only and is configured to fail fast.

PR #1174, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
Does travis_retry really solve the build issues? According to the data in paper [An empirical study of the long duration of continuous integration builds](https://dl.acm.org/doi/10.1007/s10664-019-09695-9), travis_retry can only solve 3% of the build failures. And it may cause unstable build and increase build time.

=====================


tfr42:
Thank you for your contribution. Unfortunately, after carefully reviewing your contribution we, the TMC members, came to the conclusion not to merge it into the deegree code base. 
Since we aren't using travis-ci anymore and using Jenkins CI instead (see #1079 and releated PR #1179 for more information) the proposed changes to travis config are obsolet. 
Of course this is not the final word. If you think, that we did wrong then please re-initiate the Decision Making Process by making a proposal and start a discussion on the mailing list.  Anyhow, we do appreciate your commitment to open source and to the OSGeo project deegree in special.

PR #1180, opened 2021-09-17
Implemented ApiKey/Token authentication

dstenger:
Enables authentication via X-API-Key when REST interface is used.


copierrj:
A few small remarks:
- Documentation is missing.
- Having a read-only workspace configuration implicitly disables access to the API.
- ConfigServlet.token is a static field. It should be non static and initialized in the init method.

stephanr:
I am closing this in favor of the reworked #1498

PR #1181, opened 2021-09-20
Remove deegree-sqldialect-postgis dependency from deegree-featurestore-sql

dstenger:
Removed specialized dependency from FeatureStore module.


tfr42:
Build job fails https://buildserver.deegree.org/view/All/job/deegree-3.4/job/pull-request/view/change-requests/job/PR-1181/

PR #1192, opened 2021-09-29
Fix rendering of geometries in deegree

dstenger:
This PR makes `miterLimit` configurable.
When the configured factor is exceeded portrayal changes from JOIN_MITER to JOIN_BEVEL (see https://docs.oracle.com/javase/tutorial/2d/geometry/strokeandfill.html).

Configuration can be set via JNDI-Env or System-Property.

Example:
```
<Environment name="deegree.rendering.stroke.miterlimit" value="2.66" 
                       type="java.lang.Float" override="false" 
                       description="deegree Rendering - Miter Limit Factor"/>
```
or
```
-Ddeegree.rendering.stroke.miterlimit=2.66
```


tfr42 (commented):
See comment on PR #1287 about missing documentation.

stephanr:
> Rename the type Tunable to TunableParameter

This was done in the follow up PR #1330

PR #1194, opened 2021-09-30
Fix rendering of dashed polygons

dstenger:
This pull request adds ScgImageTranscoder to be able to use svg for graphic fills.


tfr42:
Relates to #1405 and #1410.

PR #1196, opened 2021-09-30
Fix serveral bugs in REST API

dstenger:
This pull request bases on https://github.com/deegree/deegree3/pull/938.

Several bugs discovered in restart and updatebboxes functionalities of REST interface were fixed and error handling was improved.


deegree-services/deegree-services-config/pom.xml
```diff
@@ -17,7 +17,7 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.deegree</groupId>
+      <groupId>${project.groupId}</groupId>
```
stephanr, line 20: Please align this to the deegree codebase and use the groupId without variable.

dstenger: Solved in PR #938.

tfr42:
Related to issue #1328.

Issue #1140, opened 2021-04-20
Constraint ManageStoredQueries is always set to true


dstenger:
Constraint ManagedStoredQueries in capabilities document is always set to true and all ManagedStoredQueries operations are listed in the capabilities document even if directory `storedqueries/managed` does not exist in the workspace which deactivates ManagedStoredQueries functionality.

So, although the capabilities document states that ManagedStoredQueries are supported (but the workspace configuration says otherwise), the service throws following exception when trying to create a StoredQuery:
```
<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="2.0.0"
                                 xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd"> <ows:Exception
                                 exceptionCode="OperationProcessingFailed"> <ows:ExceptionText>Performing CreateStoredQuery requests is not configured.</ows:ExceptionText>
                                 </ows:Exception> </ows:ExceptionReport>
```

Expected behavior is that constraint ManagedStoredQueries is set to false and ManagedStoredQueries operations are not listed in the capabilities document when folder `storedqueries/managed` does not exist.

PR #1212, opened 2021-10-12
fix: GetCapabilities reflect if managing stored queries is enabled

stempler:
Managing stored queries is enabled based on the presence of a folder in
the workspace. This was not reflected in the capabilities, which always
stated that managing stored queries is supported.


stempler:
Related issue was #1140 

A PR with similar changes (#1147) was merged now.

PR #1213, opened 2021-10-15
Add data to utah workspace

dstenger:
Data in utah workspace were missing.
They are added via this pull request.


tfr42:
Currently the raster files and shape files are missing.

PR #1216, opened 2021-10-21
Add check of transparent parameter

### LINKED ISSUE #1184: deegree WMS 1.3 does not return exception when transparent parameter contains invalid value
This occurs when a GetMap request is send:
* https://inspire.brandenburg.de/services/ef_sem_wms?BBOX=228152.00000000%2C5690412.00000000%2C493382.00000000%2C5939023.00000000&CRS=EPSG%3A25833&FORMAT=image%2Fpng&HEIGHT=500&LAYERS=EF.EnvironmentalMonitoringProgrammes&REQUEST=GetMap&SERVICE=WMS&STYLES=EF.EnvironmentalMonitoringProgrammes.Default.Point&TRANSPARENT=zzzz&VERSION=1.3.0&WIDTH=500

Chapter `7.3.3.9 TRANSPARENT` of specification (OGC® 06-042) defines the following:
```
The optional TRANSPARENT parameter specifies whether the map background is to be made transparent or not. 
TRANSPARENT  can  take  on  two  values,  "TRUE"  or  "FALSE".  The  default  value  is  FALSE  if  this  parameter  is  
absent from the request.
...
```

Chapter `6.11  Service exceptions` of specification (OGC® 06-042) says:
```
Upon receiving a request that is invalid according to this International Standard, the server shall issue a service 
exception report. The service exception report is meant to describe to the client application or its human user the 
reason(s) that the request is invalid.
...
```

It is not 100% clear to me, but I assume that an exception is correct in this case.
Should be discussed during TMC Meeting.

dstenger:
Fixes #1184.

For details, please see the referenced issue.

### DIFF
--- a/deegree-core/deegree-core-protocol/deegree-protocol-wms/src/main/java/org/deegree/protocol/wms/ops/GetMap.java
+++ b/deegree-core/deegree-core-protocol/deegree-protocol-wms/src/main/java/org/deegree/protocol/wms/ops/GetMap.java
@@ -140,7 +140,7 @@ public class GetMap extends RequestBase {
     /**
      * @param map
      * @param version
-     * @param service
+     * @param exts
      * @throws OWSException
      */
     public GetMap( Map<String, String> map, Version version, MapOptionsMaps exts ) throws OWSException {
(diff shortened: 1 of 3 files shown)


copierrj:
The TMC decided that changes to the protocol handling should obey the 'strict' toggle of the service. Please amend this PR accordingly.

lgoltz:
> The TMC decided that changes to the protocol handling should obey the 'strict' toggle of the service. Please amend this PR accordingly.

PR was enhanced: https://github.com/deegree/deegree3/pull/1216/commits/f320dde4f6363179bf9365b1e30ef70591920a70

PR #1238, opened 2021-12-10
Implement ExtendedDescription for WFS 2.0

dstenger:
ExtendedDescription was implemented for WFS 2.0 as specified in http://schemas.opengis.net/wfs/2.0/wfs.xsd.


tfr42 (changes requested):
License header are missing

copierrj:
WmsCapabilities111ThemeWriterTest uses the org.xmlmatchers package which doesn't seem to on the classpath.

PR #1261, opened 2022-01-06
Improve MAVEN build Performance

xiayingfeng:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

=====================


copierrj:
Thank you for your contribution. Unfortunately, after carefully reviewing your contribution we, the TMC members, came to the conclusion not to merge it into the deegree code base because this change would complicate our current CI setup. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #1266, opened 2022-01-14
Improve MAVEN build Performance

xiayingfeng:
According to [Maven parallel test](https://www.baeldung.com/maven-junit-parallel-tests), we can run tests in parallel.

That report generation takes time, slowing down the overall build. Reports are definitely useful, but do you need them every time you run the build. We can conditionally disable generating test reports by setting `<disableXmlReport>true<disableXmlReport>`. If you need to generate reports, just add `-DcloseTestReports=false` to the end of mvn build command.

=====================

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1089,6 +1089,7 @@
     <spring-framework.version>5.3.12</spring-framework.version>
     <log4j.version>2.17.1</log4j.version>
     <slf4j.version>1.7.32</slf4j.version>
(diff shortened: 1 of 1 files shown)


copierrj:
Thank you for your contribution. Unfortunately, after carefully reviewing your contribution we, the TMC members, came to the conclusion not to merge it into the deegree code base because this change would complicate our current CI setup. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #1285, opened 2022-03-11
Allow to specify custom dialect provider on a data source level

stephanr:
This enhancement allows to either override the automatic SQL dialect provider selection, for customized dialects, or to specify the correct dialect if deegree cannot determine the correct provider based on the JDBC URL.


stephanr:
Last build (4) ran without issue see https://buildserver.deegree.org/blue/organizations/jenkins/deegree-3.4%2Fpull-request/detail/PR-1285/4/pipeline/

PR #1287, opened 2022-03-11
Fix geometry conversion for simple point geometries on oracle databases

stephanr:
Until now, the geometry would have never been stored as an optimized point for a 2D point.
This originates from the check for `ordinates.length > 2 && ordinates.length < 4` which only applies to 3D points at the moment.

To fix this bug and also to prevent Multipoint (2005/3005) geometries with only a single point to be stored wrongly, the type check was narrowed down to 2D/3D simple point (2001/3001).

It also may or may not be a desired option to never use optimized point storage, so the option to use `-Ddeegree.sqldialect.oracle.optimized_point_storage=false` was included.


stephanr:
Last build (4) ran without issue see https://buildserver.deegree.org/blue/organizations/jenkins/deegree-3.4%2Fpull-request/detail/PR-1287/4/pipeline

stephanr:
I will change the System Property Usage and Documentation introduced in #1192

tfr42 (commented):
Documentation of the system property should be added.

PR #1288, opened 2022-03-11
Add an extension to the oracle dialect that works analog to the FeatureInspector

stephanr:
This Pull Request contains a new interface `org.deegree.sqldialect.oracle.sdo.SDOInspector` that can be used to interfere before a geometry is written to or after the geometry has been read from the database.
With this `org.deegree.sqldialect.oracle.sdo.SDOInspector`, which works similar as `org.deegree.feature.persistence.FeatureInspector`, it is possible to handle custom geometries and or specialized simplification with an extended dialect.

This PR is logically related to PR #1285


stephanr:
After the problems that appeared after the rebase were solved, the build was successfully.
See https://buildserver.deegree.org/blue/organizations/jenkins/pull-request/detail/PR-1288/39/pipeline/

PR #1291, opened 2022-03-11
Acceleration of batch import jobs by enabling the use of sql batch operations

stephanr:
To improve batch loading speed, it is critical to reduce the amount of database round trips for each operation.
To achieve this, it is required to prevent auto generated columns when `idGenMode == IDGenMode.USE_EXISTING && idGenMode.isSkipResolveReferences()` is used and all identifying column values are already known.

In this PR the handling of auto generated columns was optimized in `InsertRow` and also support for batching was introduced in `InsertRowManager`.


stephanr:
Last build (4) ran without issue see https://buildserver.deegree.org/blue/organizations/jenkins/deegree-3.4%2Fpull-request/detail/PR-1291/4/pipeline

stephanr:
At today's tmc meeting (see https://github.com/deegree/deegree3/wiki/tmc-meeting-minutes-20221221),  we discussed this pull request and concluded that it is too specific to be included in its current form.
However, it would still be appreciated to keep opther fixes which are not related to SQL batching.

stephanr:
This PR will be closed for now

PR #1307, opened 2022-04-13
Fix packaging of log4j2.properties file in deegree WAR file

### LINKED ISSUE #1306: Log4j configuration files do not exist in deegree 3.4.17+ versions
The Log4j configuration files of deegree versions 3.4.17 and higher do not exist. The path 'deegree-webservices/WEB-INF/classes of a deployed webapp' is therefore empty. Versions of 3.4.16 and lower still have the configuration files in the above mentioned filepath.

tfr42:
Fixes #1306  packaging of log4j2.properties file in WAR file in WEB-INF/classes and not in JAR file.


tfr42:
Replaced by PR #1312

PR #1315, opened 2022-05-02
fix documentation on DisableDynamicSchema

stempler:
As the description already states, dynamic schema generation usually has to be disabled for complex application schemas.
But to disable, the option needs to be set to `true` and not to `false`.


julianzz98 (approved):
The behavior was tested in deegree with an INSPIRE workspace and can be confirmed to be as the PR is proposed.

PR #1330, opened 2022-05-12
Fix rendering of geometries in deegree

stephanr:
Replacement of PR #1192
Original description:

This PR makes `miterLimit` configurable.
When the configured factor is exceeded portrayal changes from JOIN_MITER to JOIN_BEVEL (see https://docs.oracle.com/javase/tutorial/2d/geometry/strokeandfill.html).

Configuration can be set via JNDI-Env or System-Property.

Example:
```
<Environment name="deegree.rendering.stroke.miterlimit" value="2.66" 
                       type="java.lang.Float" override="false" 
                       description="deegree Rendering - Miter Limit Factor"/>
```
or
```
-Ddeegree.rendering.stroke.miterlimit=2.66
```

### DIFF
--- a/deegree-services/deegree-webservices-handbook/src/main/asciidoc/appendix.adoc
+++ b/deegree-services/deegree-webservices-handbook/src/main/asciidoc/appendix.adoc
@@ -3,7 +3,7 @@
 
 The following chapters of the documentation are aimed at users with specialized knowledge on the inner workings of deegree.
 
-NOTE: The functionality is currently in evolving status and may be change without future notice.
+NOTE: As the internal workings of deegree are always evolving, the information in these chapters might become obsolete without any prior notice. There is no guarantee for anything described in here to be the same for future versions of deegree.
 
 === Tunable deegree parameters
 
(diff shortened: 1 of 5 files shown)


deegree-services/deegree-webservices-handbook/src/main/asciidoc/appendix.adoc
```diff
@@ -0,0 +1,38 @@
+[[anchor-appendix]]
+== Appendix
+
+The following chapters of the documentation are aimed at users with specialized knowledge on the inner workings of deegree.
+
+NOTE: The functionality is currently in evolving status and may be change without future notice.
```
copierrj, line 6: Suggestion for a more explicit wording of this note:

As the internal workings of deegree are always evolving, the information in these chapters might become obsolete without any prior notice. There is no guarantee for anything described in here to be the same for future versions of deegree.

tfr42: +1

PR #1345, opened 2022-06-09
Fix default values of maxFeatures and featureInfoRadius

dstenger:
This pull request fixes the default values of maxFeatures and featureInfoRadius.
Previously, when requesting a WFS without maxFeatures, no features were returned as 0 was set for this parameter.
This PR fixes that by setting -1 default values.


tfr42:
Regression introduced with PR #700.

tfr42 (approved):

PR #1353, opened 2022-06-27
Implement PointsList#copy() and JTSPoints#copy()

dstenger:
Interface org.deegree.geometry.primitive.Point was enhanced by copy() method.

Afterwards, PointReference and DefaultPoint were enhanced by implementations of this new method.

Finally, PointsList#copy() could be implemented.
JTSPoints#copy() was also implemented.

### DIFF
--- a/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/tom/Reference.java
+++ b/deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/tom/Reference.java
@@ -79,6 +79,15 @@ public class Reference<T extends Object> implements Object {
     }
 
     /**
+     * Returns the resolver.
+     *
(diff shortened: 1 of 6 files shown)


tfr42 (changes requested):
Revisit the implementation blocks for clone() and copy() if they can be adapted.

stephanr:
During the TMC meeting the question came up if the introduction of copy() is useful while the existing API already uses clone(). (And copy() seems to come from a JTS interface in https://github.com/deegree/deegree3/blob/main/deegree-core/deegree-core-geometry/src/main/java/org/deegree/geometry/points/Points.java)

stephanr (approved):

tfr42 (approved):

PR #1367, opened 2022-07-27
Upgrade to PostGIS 2.5.1, usage of JTS WKTReader instead of PostGIS WKTReader

julianzz98:
Upgrade to PostGIS 2.3.0 to 2.5.1 and switch from PostGIS WKTReader to JTS WKTReader.


julianzz98 (commented):
Sorry for merge-commit @copierrj

tfr42 (changes requested):
The proposal is to move the method createFromJTS() to GeometryUtils as a static method if possible to avoid object creation for converting objects.

tfr42 (approved):

PR #1204, opened 2021-10-04
Enhance SLD/SE functions

dstenger:
This pull request contains multiple SLD/SE enhancements.

This includes:
* SPI for Graphic/Mark/WellKnownName
* Optimization of fill
* Function AchoredSymbol
* WKTParser2
* Enhancement of WKTWriter 
* GeoServer symbols ( shape://, extshape://, qgis:// )
* TTF symbol syntax with GeoServer syntax (ttf://)
* Symbols basing on WKT syntax (wkt://)
* SVG path syntax (svgpath://)
* ogc:Function HatchingDistance

Some tests are not passing and are ignored: https://github.com/lat-lon/deegree3/blob/enhancement/migrations-styling-enhacments-dee-83-7692/deegree-core/deegree-core-geometry/src/test/java/org/deegree/geometry/WKTReader2JTSGeometryTests.java
Those tests should be fixed before merging this pull request.

This PR includes PR  #1244 (JTS).

PR #1407, opened 2022-09-30
Add support for rectangular (box-styled) halo

stephanr:
This PR is a (smal) reworked part of the larger PR #1204

It adds the possibility to use rectangular helo effects when rendering text.
This is similar to the way deegree2 rendered halo as boxes.

It also contains an edge-case fix so that when halos are set (<> 0) it will be rendered with at least a size of one pixel.

**EDIT:**
For the test to work, it was necessary to include and load a font so that the test was run with the same font.
The Fonts distributed with each JDK distribution differ and can as well differ on the platforms (win/mac/linux).

**EDIT2:**
The previously added Font was replaced with an maven dependency.
The Font is licensed, under "SIL OPEN FONT LICENSE Version 1.1" see also https://search.maven.org/artifact/com.mpobjects.jasperreports.font/jasperreports-fonts-liberation/2.1.2/jar


stephanr:
Fixed merge conflict by rebase.

deegree-services/deegree-webservices-handbook/src/main/asciidoc/renderstyles.adoc
```diff
@@ -874,7 +874,7 @@ The first parameter is the hatching angle, the second is the line spacing in the
 </PolygonSymbolizer>
 ----
 TIP: For of the shelf hates, which will create nice results, use the mark symbol `shape://slash`, `shape://backslash`
-or `shape://times` for  45°, `shape://horline` for 0° and `shape://vertline` for 90° hatches.
+or `shape://times` for  45�, `shape://horline` for 0� and `shape://vertline` for 90� hatches.
```
tfr42, line 877: Please review the encoding of this file. The special character for deegree is not displayed correctly.

tfr42 (changes requested):
Please check if the  TTF files may be added via a maven dependency.

deegree-services/deegree-webservices-handbook/src/main/asciidoc/renderstyles.adoc
stephanr: The encoding was fixed in https://github.com/deegree/deegree3/pull/1407/commits/ec252f76e0908acc3bc94ceb3cc28419c74c2376

tfr42 (approved):

PR #1410, opened 2022-10-10
Extending the SLD/SE rendering capabilities of' Mark for more symbols and better area fill

stephanr:
This PR is the, **reworked**, main part of the larger PR #1204 

The following part are included:

- a new extension point, which allows implementing custom Graphic/Mark/WellKnownName handlers.
- (new) test cases which render a feature collection with a style and compare the results to the image
  -  bbox is provided by the featurecollection, with and height is taken from the reference image
- corrections that improve the rendering of symbols for area fills
- documentation is reduced to the parts included in this pr
- the implementation for `WellKnownName` whit the prefixes `shape://`, `extshape://`, `qgis://` and `ttf://` do not include new or changed dependecies and are reworked versions from their geoserver/geotools counterpart (which is also LGPL 2.1 or newer)
  - see [extshape://](https://github.com/geotools/geotools/blob/27.0/modules/library/render/src/main/java/org/geotools/renderer/style/markwkt/MeteoMarkFactory.java) [shape://](https://github.com/geotools/geotools/blob/27.0/modules/library/render/src/main/java/org/geotools/renderer/style/ShapeMarkFactory.java) [qgis://](https://github.com/geotools/geotools/blob/27.0/modules/library/render/src/main/java/org/geotools/renderer/style/QGISMarkFactory.java) [ttf://](https://github.com/geotools/geotools/blob/27.0/modules/library/render/src/main/java/org/geotools/renderer/style/TTFMarkFactory.java) (The references are to a newer version of geotools)
- include a helper function to calculate the size of an mark for simplified hatching


stephanr:
The PR was rebased to the current main, and also the obsolete change of ShapeStroke was removed.

PR #1411, opened 2022-10-10
Extending the SLD/SE rendering capabilities of Mark for symbols based on SVG paths

stephanr:
This PR is an extension to #1410 

This adds the implementation of `WellKnownName` for  `svgpath://` which generates shapes from SVG path syntax [1].
It reuses the dependency of batik.

The only add-on to the base PR is the commit https://github.com/deegree/deegree3/pull/1411/commits/00c47531e59c67a0056c7378ce9bf8020c14376b 

[1] https://www.w3.org/TR/SVG/paths.html#PathData


stephanr:
Rebased onto current version of #1410

PR #1424, opened 2022-11-11
Fix conversation of foot within unit of measure handling

stephanr:
Values given in foot where previously handled incorrectly like meter.
Now uses the conversation factor of 0.3048 meter per foot as defined at the [International Yard and Pound Agreement](https://en.wikipedia.org/wiki/International_Yard_and_Pound_Agreement) in 1959.

Source: http://en.wikipedia.org/wiki/Foot_%28unit%29


tfr42:
If more conversions of this kind are needed we may consider to use [javax.measure](https://jcp.org/en/jsr/detail?id=363).

PR #333, opened 2014-07-30
OGC GeoPackage implementation - 3.4

diegomigliavacca:
http://tracker.deegree.org/deegree-services/ticket/613

"The GeoPackage specification describes an open, standards-based, platform-independent, portable, self-describing, compact format for transferring geospatial information. It is a set of conventions for SQLite to store interoperable Features and/or Tiles on a common base" (opengis/geopackage GitHub repository).

For the feature part, the user needs to configure in Deegree:
an SQL feature store;
a WFS service;
a JDBC connection;
a Feature Layer.

For the tile part, the user needs to configure:
a WMTS service;
a Gpkg Tile Store;
a Tile Layer.

Configuration of a JDBC connection (example):

```
<JDBCConnection configVersion="3.0.0" xmlns="http://www.deegree.org/jdbc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="http://www.deegree.org/jdbc http://schemas.deegree.org/jdbc/3.0.0/jdbc.xsd">

 <Url>jdbc:sqlite:/home/user/World.gpkg</Url>
 <User>null</User>
 <Password>null</Password>

</JDBCConnection>
```

User and Password can just to be set to "null", given that GeoPackage is based on SQLite.

Configuration of a Gpkg Tile Store (example):

```
<GpkgTileStore xmlns="http://www.deegree.org/datasource/tile/gpkg" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.deegree.org/datasource/tile/gpkg http://schemas.deegree.org/datasource/tile/gpkg/3.2.0/geopackage.xsd" configVersion="3.2.0">

  <TileDataSet>
    <Identifier>gpkg_tiledataset</Identifier>
    <File>home/user/World.gpkg</File>
    <TileMapping table="fromosm_tiles"/>
    <!-- [0..1]: the mime type of the desired image output format.   Default is image/png -->
    <ImageFormat>image/png</ImageFormat>
  </TileDataSet>

</GpkgTileStore>
```

PR #1437, opened 2022-12-08
Add support for GeoPackage to deegree

tfr42:
With this PR the support of GeoPackage is added to deegree. This PR is an update of the source PR  #333 (developed by @diegomigliavacca) with changes to be able to merge into the 3.5 main branch. 
It has been tested with the workspace available here: https://github.com/lat-lon/deegree-workspaces/tree/gpkgWorkspace-8209

When merging this PR consider to make a squash and merge commit.


tfr42:
Proposal is to rename the dialect to SqliteDialect instead of GpkgDialect.

stephanr:
As discussed in the tmc meeting, the documentation should mention the native reference to the native dependencies.
See also https://github.com/xerial/sqlite-jdbc#how-does-sqlitejdbc-work

PR #1446, opened 2022-12-30
Re-activate the remote WMS integration tests

### LINKED ISSUE #1246: Fix and enable deegree-wms-remotewms-tests
Pull Request #1245 enables deegree-wms-remotewms-tests module again. However, the containing tests are ignored.

Thus, the tests of deegree-wms-remotewms-tests module shall be fixed and enabled again.

tfr42:
Fixes #1246.

This PR re-activates the tests in module `deegree-wms-remoteows-tests` executed as part of the integration tests.


deegree-core/deegree-core-commons/src/main/java/org/deegree/commons/utils/test/IntegrationTestUtils.java
stephanr, line 1: I think the base64 encoded output to STDOUT was intentional, because by default the LOG is very strongly suppressed on the CI/CD system.
Had you checked whether the results still appear in the CI/CD log after the change in case of an image error ?

PR #1448, opened 2022-12-30
Improved documentation how to configure deegree for offline startup with schema-driven feature stores

### LINKED ISSUE #1066: Add support for offline xml schema bundles
Many deegree users use externally hosted xml schema's in their setup. Unfortunately this is causing problems when this service suddenly disappears resulting in a deegree instance that is no longer able to boot. Therefore deegree should support using an offline bundle of these external resources. These locally available files are to be used instead of doing http(s) request to another service.

tfr42:
Fixes #1066.

Update documentation how to retrieve and store schema files locally when using schema-driven SQLFeatureStore configurations.


tfr42:
New OGC schemas file is available https://repo.deegree.org/service/rest/repository/browse/releases/org/deegree/deegree-ogcschemas/20230513/

stephanr (approved):

stephanr (approved):

PR #1452, opened 2023-01-13
Enforce UTF-8 as platform encoding in unit tests

KevinvdBosch:
Without setting the encoding explicitly some unittest fail on systems that do not use UTF-8 as platform encoding.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,7 +140,7 @@
           <version>3.0.0-M7</version>
           <configuration>
             <!-- Travis build workaround -->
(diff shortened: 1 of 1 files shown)


tfr42 (approved):

dstenger (approved):

tfr42:
@KevinvdBosch Thank you for providing this change. We will accept and merge it. For future contributions to the deegree project please accept the deeegree committer guidelines as documented here: https://github.com/deegree/deegree3/blob/main/CONTRIBUTING.md#legal

PR #1457, opened 2023-01-20
JAXB v3

KevinvdBosch:
This pull request contains changes from using jaxb v2 to v3. This means converting the use of javax packages to jakarta.

The only problem with this is that xsd files can't be found anymore when they are in separate modules. It seems that maven can't merge them properly anymore.

A temporary fix could be to duplicate the xsd files where needed. This will make all unit tests pass.

The only thing that still remains broken is the creation of the zip file containing all schema files in the deegree-core-schema module.

depends on #1452

### DIFF
--- a/deegree-core/deegree-connectionprovider-datasource/pom.xml
+++ b/deegree-core/deegree-connectionprovider-datasource/pom.xml
@@ -18,8 +18,8 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.jvnet.jaxb2.maven2</groupId>
-        <artifactId>maven-jaxb2-plugin</artifactId>
+        <groupId>com.evolvedbinary.maven.jvnet</groupId>
+        <artifactId>jaxb30-maven-plugin</artifactId>
       </plugin>
     </plugins>
   </build>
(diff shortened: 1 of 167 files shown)


stephanr:
> The only problem with this is that xsd files can't be found anymore when they are in separate modules. It seems that maven can't merge them properly anymore.

Could this be related to our custom resolver configured here: 

https://github.com/deegree/deegree3/blob/86bbe9165172fea0f1ef8c2e701d5eeebcdfa941/pom.xml#L97

Have you checked whether this interface also needs to be changed?
https://github.com/deegree/deegree3/blob/86bbe9165172fea0f1ef8c2e701d5eeebcdfa941/uncoupled/deegree-jaxb-resolver/src/main/java/org/deegree/uncoupled/jaxb/CatalogResolver.java#L69

tfr42:
Hi @KevinvdBosch, thank you for providing that PR. 
To make it easier to merge the proposed changes would it be possible that you
- split the PR related to JAXB 2 to 3 migration from
- changing old JAVA EE API like javax.mail and javax.activation (imports and dependencies)?
- And also create a seperate PR for changing `com.sun.xml.bind:jaxb-impl` to `org.glassfish.jaxb:jaxb-runtime` (please see #1459 for more information)

PR #1472, opened 2023-02-20
Update the schemaLocation URL in config templates to be valid

julianzz98:
The schemaLocation of the example templates of the database connectionprovider were malformed and should be functional again.


stephanr:
On this topic is the documentation here also incorrect ?

https://github.com/deegree/deegree3/blob/6ed6ffaff6f22b7e9df992d4134692fa48578b5a/deegree-services/deegree-webservices-handbook/src/main/asciidoc/serverconnections.adoc#minimal-configuration-example-postgresql

julianzz98:
Thanks for the hint, the documentation will be revised in a following commit!

tfr42 (approved):

tfr42 (approved):

Issue #824, opened 2017-04-25
bbox_cache.properties contains outdated bounding boxes


lgoltz:
The bounding box containing all features of a feature type is stored in the bbox_cache.properties if a SQLFeatureStore store is used. 
If a feature is inserted the cached bounding box is extended if required. But a delete or update transaction does not  lead to a smaller bbox.

In the method updating the BBoxCache [1] a TODO can be found and a debug message is written:
```
// TODO configuration switch for bbox recalculation strategy
if ( !recalcFTs.isEmpty() ) {
  LOG.debug( "Full recalculation of feature type envelopes required. Delete 'bbox_cache.properties' if you need minimal envelopes." );
}
```
[1] org.deegree.feature.persistence.sql.SQLFeatureStoreTransaction#updateBBoxCache

lgoltz:
Related issue: #562

PR #1477, opened 2023-03-08
Add bbox_cache.properties per SQLFeatureStore

julianzz98:
This PR replaces the current bbox_cache.properties with bbox_cache_<SQLFeatureStoreId>.properties, therefore enabling the configuration of an individual spatial extend of FeatureTypes with the same name configured in different SQLFeatureStore.

This PR does not solve issue #824!

### DIFF
--- a/deegree-datastores/deegree-featurestores/deegree-featurestore-commons/src/main/java/org/deegree/feature/persistence/FeatureStoreManager.java
+++ b/deegree-datastores/deegree-featurestores/deegree-featurestore-commons/src/main/java/org/deegree/feature/persistence/FeatureStoreManager.java
@@ -41,9 +41,6 @@
  ----------------------------------------------------------------------------*/
 package org.deegree.feature.persistence;
 
-import java.io.File;
-import java.io.IOException;
-
 import org.deegree.feature.persistence.cache.BBoxCache;
 import org.deegree.feature.persistence.cache.BBoxPropertiesCache;
 import org.deegree.workspace.Workspace;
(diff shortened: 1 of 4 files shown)


deegree-datastores/deegree-featurestores/deegree-featurestore-commons/src/main/java/org/deegree/feature/persistence/FeatureStoreManager.java
```diff
@@ -53,43 +50,58 @@ Occam Labs UG (haftungsbeschränkt)
...
-            // else?
         } catch ( IOException e ) {
-            LOG.error( "Unable to initialize global envelope cache: " + e.getMessage(), e );
+            LOG.error( "Unable to initialize envelope cache for feature store with id " + featureStoreId + ": "
+                       + e.getMessage(), e );
         }
-        super.startup( workspace );
-    }
-
-    public BBoxCache getBBoxCache() {
```
stephanr, line 91: Should this method be deleted, as it is now breaking implementations which reley on it.

tfr42:
Add information about the change to the configuration in https://github.com/deegree/deegree3/wiki/Upgrading-deegree-webservices

julianzz98:
According to the TMC, a use case should be added for better understanding of the PR. Furthermore the wiki should be updated regarding breaking changes in deegree 3.5-SNAPSHOT.

tfr42:
The current existing behaviour shall be kept. The new behaviour shall be only applied if a file with the proposed name such as `bbox_cache_FEATURE_STORE_ID.properties` is present.

stephanr (approved):

tfr42 (approved):

stephanr (approved):

Issue #1260, opened 2022-01-05
Reevaluate the code style used for deegree


tfr42:
The code style https://github.com/deegree/deegree3/wiki/Developer-guidelines#source-code-formatting-in-eclipse needs to be reevaluated how it is applicable using current IDE's. 

PR shall respect the current code style, and documented in https://github.com/deegree/deegree3/wiki/Working-with-pull-requests#checklist-for-pull-requests

tfr42:
Another approach would be to use maven plugins to apply the formatting rules. 
Available Maven Plugins:
- https://github.com/diffplug/spotless/tree/master/plugin-maven
- https://github.com/spring-io/spring-javaformat

PR #1507, opened 2023-05-13
Remove maven-checkstyle-plugin

dstenger:
Removed maven-checkstyle-plugin as it is not active by default and there are multiple failures when executing it via `mvn checkstyle:check`.

Will be replaced by spring-javaformat plugin (see #1260).


tfr42 (approved):
Will be replaced by #1260

stephanr (approved):

PR #1519, opened 2023-05-17
Fixed build step for quality checks by using the warnings ng jenkins plugin

tfr42:
fixes #1416 for 3.5.x line

PR #1528, opened 2023-06-05
Enhanced pipeline to publish artifacts on jenkins

tfr42:
This PR enhances the pipeline to re-enable the publishing of artifacts created during install **and** release step.
Reverting the change introduced with #1513.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,7 +63,7 @@ pipeline {
         }
         stage ('Release') {
             when {
-                allOf{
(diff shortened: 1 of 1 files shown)


stephanr:
I think this change should reference #1519

tfr42:
The PR shall be enhanced to archive also all artifacts created by the multibranch pipeline step "test".

stephanr (approved):
looks good to me

PR #1534, opened 2023-06-21
Fix transitive dependency to jai_core to use jai-core instead

tfr42:
This PR excludes the `jai_core` dependency to resolve issues when building deegree and `jai_core` instead of `jai-core` is used. **Note:** within the project `jai-core` should be used as artifactID.
This issue occurs occasionally when building deegree with:  `mvn -T6 clean install` or different order when defining `-pl` option.


tfr42:
Note: Check the enforcer `ignoreClasses` if `com.sun.media.jai.*`.

PR #1550, opened 2023-07-26
Use targed gml namespace if boundedBy/geometry is in gml namespace

### LINKED ISSUE #1548: WFS 1.1.0 GetFeature Request fails with BLOB Mode
The GetFeature request http://localhost:8080/deegree-webservices/services/wfs?request=GetFeature&Service=WFS&Version=1.1.0&typename=test fails with 

```
<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows http://schemas.opengis.net/ows/1.0.0/owsExceptionReport.xsd" version="1.0.0">
   <ows:Exception exceptionCode="NoApplicableCode">
    <ows:ExceptionText>Unbound namespace URI 'http://www.opengis.net/gml'</ows:ExceptionText>
  </ows:Exception>
</ows:ExceptionReport>
```

A simple WFS is configured with SQLFeatureStore in BLOB Mode.  

The request fails in current 3.5.1-SNAPSHOT as well as in 3.4.33 with the same failure.

dstenger:
Fixes #1548
Please also see this issue for description of the bug.

The problem is that in Blob mode the GML version of the read features is used (e.g. GML 3.2). This version is also used for the response although another GML version is expected (e.g. GML 3.1) leading to the documented exception.

This PR enables that the targed GML namespace is used if boundedBy/geometry is in GML namespace preventing the mixing of GML versions.

### DIFF
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/feature/GMLFeatureWriter.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/feature/GMLFeatureWriter.java
@@ -371,7 +371,9 @@ public class GMLFeatureWriter extends AbstractGMLObjectWriter {
 				writeNilledElement(propName, attributes);
 			}
 			else {
-				writeStartElementWithNS(propName.getNamespaceURI(), propName.getLocalPart());
+				String namespaceURI = GMLSchemaInfoSet.isGMLNamespace(propName.getNamespaceURI()) ? gmlNs
+						: propName.getNamespaceURI();
+				writeStartElementWithNS(namespaceURI, propName.getLocalPart());
 				if (value != null) {
 					gmlStreamWriter.getGeometryWriter().exportEnvelope((Envelope) value);
 				}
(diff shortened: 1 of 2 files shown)


dstenger:
Another test should be done checking whether the generated GML 3.1 (from GML 3.2 Blob store) is schema valid.

stephanr:
The fix seems to be reasonable to fix the output to be valid XML again.

Currently I can not estimate what output will appear for GML3.2 blob data after the fix is applied.
Will the FeatureCollection contain a Geometry in GML 3.2 or GML 3.1 ?
Another question would be if the Capabilities for this endpoint contains both GML versions or only 3.2.

If the endpoint only shows GML3.2, or can be changed to, as supported output format in the capabilities, another approach may be to force GML3.2 output in the case of BLOB storage. 
This might be not compliant to OGC WFS 1.1.0 but at least consistent and an alternative to rewrite geometries of BLOB storage to different GML versions.

tfr42:
The description of this PR shall be extended with the information how it behaves with the fix (paste the GML output here) for the different GML versions.

stephanr (approved):

tfr42 (approved):

PR #1573, opened 2023-09-12
Fix dryRun in GmlLoader

lgoltz:
Executing the GmlLoader with the parameter `-dryRun=true` results in 

```
ERROR o.s.batch.core.job.AbstractJob - Encountered fatal error executing job

org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'scopedTarget.step' defined in org.deegree.tools.featurestoresql.loader.GmlLoaderConfiguration: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.batch.core.Step]: Factory method ‘step’ threw exception; nested exception is java.lang.IllegalStateException: ItemWriter must be provided 
```

This PR fixes the GmlLoader by adding a NullWriter.

### DIFF
--- a/deegree-tools/deegree-tools-gml/src/main/java/org/deegree/tools/featurestoresql/loader/GmlLoaderConfiguration.java
+++ b/deegree-tools/deegree-tools-gml/src/main/java/org/deegree/tools/featurestoresql/loader/GmlLoaderConfiguration.java
@@ -202,7 +202,8 @@ public class GmlLoaderConfiguration {
 		}
 
 		if (dryRun) {
-			return builder.build();
+			NullWriter nonWritingFeatureStoreWriter = new NullWriter();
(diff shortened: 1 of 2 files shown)


julianzz98:
A local test was able to prove that the fix is working:

Using the GmlLoader with parameter `-dryRun`:
```
root@1a83ab34b232:/opt# java -jar deegree-tools-gml-3.5.2-SNAPSHOT.jar GmlLoader -pathToFile=/Soil.gml -workspaceName=inspire_soil -sqlFeatureStoreId=inspire_soil_fs -disabledResources=https://inspire.ec.europa.eu/codelist/ -dryRun=true
```
Excerpt of the final output after running successfully:
```

tfr42 (approved):

PR #1630, opened 2024-01-11
Enhance configurability of WFS output format for defaults and CSV

### LINKED ISSUE #1618: Enhance configurability of WFS output format for defaults and CSV
The current configuration options for WFS output formats are difficult to use for end users.
For example, if a user does not want to use `GeoJSON` he has to configure all GML and CSV output formats manually.
Furthermore, the current CSV implementation appears to be compliant with the RFC 4180 definition of CSV which is, at least in Germany, not used in all commonly used products.
Here it is often seen that products use delimiters depending on configured system locale, for example `;`.

Currently, I am working on a PR to extend the current behavior of deegree on both topic.

To ease configuration of default formats, I would suggest adding a `DefaultFormats` option before other custom output formats.
And for the CSV output format, I like to extend the configuration options.

**In this regard, feedback on the proposed changes would be very welcome.**

--- 

### Snapshot of current draft of Handbook change:

**5.x.x. Output formats and defaults**
By default, a deegree WFS will offer GML 2, 3.0, 3.1, 3.2, GeoJSON and CSV as output formats and
announce those formats in the GetCapabilities responses (except for WFS 1.0.0, as this version of
the standard has no means of announcing other formats than GML 2).

NOTE: If custom format configurations are present in the configuration the default formats will be omitted.

To ease configuration, it is also possible to add custom format configurations in addition to the
default or a subset of the default ones.

Example of using the defaults, except the CSV and GeoJSON

```xml
...
<DefaultFormats>
<ExcludeMimeType>application/geo+json</ExcludeMimeType>
<ExcludeMimeType>*csv*</ExcludeMimeType>
</DefaultFormats>
...
```

NOTE
The exclusions are defined by specifying one or more mime types or patterns.
Simple placeholders like `?` for a single character or `*` for several characters can be
used.

**5.x.x Adding CSV output formats**

Using option element CsvFormat, it possible to enable CSV as GetFeature output format.
The CsvFormat option has the following sub-options:

|Option |Cardinality |Value |Description|
|--|--|--|--|
|`@encoding` |0..1| String |Character encoding, e.g. `UTF-8` to use instead of the default of the Java Servlet container.|
|`@columnHeaders` |0..1 |String |Defines how headers are derived from property name. Allowed values: `auto`, `short`, `prefixed`, `long`. Defaults to `auto` which chooses the shortest unique header name list built from `long` namespace with local part, `prefixed` prefix with local part, or `short` only the local part of the property name.|
|`@quoteCharacter` |0..1 |String |Single character to be used as delimiter in output. Defaults to comma (`,`).|
|`@escape` |0..1 |String |Single sign to be used as delimiter in output.|
|`@delimiter` |0..1 |String |Single sign to be used as escape character. Default to double quote (`"`)|
|`@instanceSeparator` |0..1 |String |Text used when the property with multiple instances is joined. Defaults to the pipe sign with spaces around (` \| `).|
|`@recordSeparator` |0..1 |String |Text to be used between records. Defaults to carriage return with line feed (`\r\n`).|
|`@geometries` |0..1 |Boolean |Defines if geometry columns should be included or not. Enabled by default.|
|MimeType |1..n |String |Mime types associated with this format configuration|
|ExtraColumns |0..1 |Complex |Defines if additional columns should be added to the output.|

The _ExtraColumns_ option has the following sub-options:
|Option |Cardinality |Value |Description|
|--|--|--|--|
|Identifier |0..1 |String |Name of the column to output the GML identifier. An empty value will omit this column, which is the default.|
|CoordinateReferenceSystem |0..1 |String |Name of the column to output the coordinate reference system name. Defaults to `CRS`.|

Example for CSV output format
```xml
...
<CsvFormat>
<MimeType>text/csv</MimeType>
</CsvFormat>
...
```
Example for a complex CSV output format
```xml
...
<CsvFormat columnHeaders="prefixed" delimiter=";" >
<ExtraColumns>
<Identifier>FID</Identifier>
<CoordinateReferenceSystem />
</ExtraColumns>
<MimeType>text/csv; subtype=semicolon</MimeType>
</CsvFormat>
...
```

WARNING
CSV output format is currently only implemented for
GetFeature requests with exactly one typename! Complex
attributes as well as attributes are currently not included
in the response.

stephanr:
This PR eases configuration of default formats and extends the configuration options for CSV output in WFS,

A detailed description is available in the issue #1618.

fixes #1618


tfr42:
No update required for https://github.com/deegree/deegree3/wiki/Upgrading-deegree-webservices because the current behaviour is not changed.

copierrj (approved):

tfr42 (approved):

PR #1666, opened 2024-03-14
Fixed evaluation of reference data (3.6)

lgoltz:
This PR fixes the evaluation of reference data. Type definitions are skipped to detect the correct number of features. Furthermore support of GenericFeature was added.

PR with 3.5 Backport: #1667


lgoltz:
Replaced by https://github.com/deegree/deegree3/pull/1673

Issue #1551, opened 2023-07-26
Adjust example configuration for the service metadata


julianzz98:
The example configuration for the service metadata under `deegree3/deegree-services/deegree-services-commons/src/main/resources/META-INF/schemas/services/metadata/example.xml` should be adjusted so it matches the information of the other example templates, e.g. WFS or WMS. 

Currently the TMC contact information is used, including the official email.

julianzz98:
The corresponding PR (https://github.com/deegree/deegree3/pull/1712) was already merged.

PR #1712, opened 2024-06-07
Adjust the information displayed in the service metadata example of deegree

julianzz98:
The information was changed so it matches the information of already adjusted example templates.

This PR closes issue #1551.


tfr42 (approved):

tfr42:
For fictitious purposes we should use numbers like this for the US: +1 206 555 0100 - +1 206 555 0199 or similar in the UK +44 113 496 0000 - +44 113 496 0999
For Germany there are some defined in https://www.bundesnetzagentur.de/SharedDocs/Downloads/DE/Sachgebiete/Telekommunikation/Unternehmen_Institutionen/Nummerierung/Rufnummern/Mittlg148_2021.pdf?__blob=publicationFile&v=1

julianzz98:
The phone number was changed to 0000/000000 and the email to info@example.com. This was done despite your suggestion, since all the other example templates (e.g. for the capabilties) are formatted this way.

PR #1729, opened 2024-08-12
[DRAFT] Added calculation of the bounding box across multiple geometry columns for PostgreSQL

julianzz98:
With this PR, the functionality of deegree was enhanced so the bounding box (BBOX) can be calculated across multiple geometry columns (PostgreSQL) when the REST API request `/deegree-webservices/config/update/bboxcache` is used. 

Prior to the change the functionality behind the mentioned REST API request would only consider the first fitting geometry column for its calculation, even if it was supposedly empty.


tfr42:
Since the interface of SQLDialect has been changed by adding new methods and an implementation is provided for PostGIS only it requires further testing if this change affects the other dialects.

tfr42:
PR #1731 for 3.6 line (`main` branch) has been merged.

PR #1730, opened 2024-08-13
Enhance bbox requests without geometry property to consider all geometry properties (3.5)

lgoltz:
This PR applies includes all available GeometryProperties of a FeatureType in a GetFeature request with bbox Parameter (without proeprty name), e.g. http://localhost:8081/deegree-webservices/services/inspirewfs_Gebaeude_3D_LoD2?service=WFS&request=GetFeature&version=2.0.0&srsName=urn:ogc:def:crs:EPSG::7423&typeNames=bu-core3d%3ABuilding&bbox=49.1,6.1,0,49.8,7.1,1000,urn:ogc:def:crs:EPSG::7423&namespaces=xmlns(bu-core3d,http%3A%2F%2Finspire.ec.europa.eu%2Fschemas%2Fbu-core3d%2F4.0)&count=1000

Prior to the change the first geometry column was considered.

### DIFF
--- a/deegree-core/deegree-core-sqldialect/deegree-sqldialect-commons/src/main/java/org/deegree/sqldialect/filter/AbstractWhereBuilder.java
+++ b/deegree-core/deegree-core-sqldialect/deegree-sqldialect-commons/src/main/java/org/deegree/sqldialect/filter/AbstractWhereBuilder.java
@@ -668,8 +668,24 @@ public abstract class AbstractWhereBuilder {
 	 * @throws FilterEvaluationException if the filter contains invalid
 	 * {@link ValueReference}s
 	 */
-	protected abstract SQLOperation toProtoSQL(SpatialOperator op)
-			throws UnmappableException, FilterEvaluationException;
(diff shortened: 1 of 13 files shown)


stephanr:
See comments on 3.6 version https://github.com/deegree/deegree3/pull/1732#issuecomment-2721297306 and https://github.com/deegree/deegree3/pull/1732#issuecomment-2722207233

copierrj:
After some discussion the TMC concluded that this change in behavior (= how deegree deals with multiple geometry attributes) is sensible but might result in surprising (to the user) results. Therefore we would like to put this change behind a tunable parameter.

In this backport we would expect the default value of this tunable to be false.

stephanr:
Thank you for your contribution. Unfortunately, after carefully reviewing your contribution we, the TMC members, came to the conclusion not to merge it into the deegree code base. Of course this is not the final word. If you think, that we did wrong then please re-initiate the decision making process by making a proposal and start a discussion on the mailing list. Perhaps we can reach a consensus then. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #1744, opened 2024-09-24
Fixed value of nilReason (3.5)

lgoltz:
nilReason was previously true:
```
<ef:operationalActivityPeriod xsi:nil="true" nilReason="true"/>
```
With the fix the value from the database is used:
```
<ef:operationalActivityPeriod xsi:nil="true" nilReason="other:unpopulated"/>
```


stephanr:
PR for 3.6 see https://github.com/deegree/deegree3/pull/1745

Issue #1718, opened 2024-06-18
Clarify the use of Spring BOM in deegree root POM dependencyManagement section


tfr42:
Currently the Spring Boot BOM is used in the project root POM in the dependencyManagement section (see https://github.com/deegree/deegree3/blob/main/pom.xml#L1206) which will import the versions defined by the Spring Boot BOM. This may override versions used in modules such as the [deegree-webservices](https://github.com/deegree/deegree3/blob/main/deegree-services/deegree-webservices/pom.xml) which uses the dependencies also defined in the [dependencyManagement section of the root project POM](https://github.com/deegree/deegree3/blob/main/pom.xml#L434).

Proposal:
- Move the Spring Boot BOM into the sub-module https://github.com/deegree/deegree3/tree/main/deegree-tools/deegree-tools-gml/pom.xml where Spring Boot is used.

stephanr:
I also think it is not a good idea to mix the `spring-boot-dependencies` BOM with the deegree dependencies, as there are overlaps between both dependency lists.

Moving the BOM import into the `deegree-tools-gml` submodule will certainly move the issue regarding overlapping dependency declaration and the resulting dependency issues into the submodule and won't interfere with most of the modules anymore. But this won't solve issues in that module if `spring-boot-dependencies` and the root pom will define different versions for dependencies (overlaps).

An alternative proposal would be to **not** import  `spring-boot-dependencies`  anymore and instead use only the BOM's which are limited to spring dependencies (for example 1, 2) and define required external dependencies with dependency management as we do with all other libs as well.
The only downside is, that we have to ensure that the overlapping dependencies will not diverge too much from the spring declared ones (which can be lookup up in `spring-boot-dependencies`).

  1. https://search.maven.org/artifact/org.springframework/spring-framework-bom/6.1.9/pom
  1. https://search.maven.org/artifact/org.springframework.data/spring-data-bom/2024.0.1/pom

PR #1746, opened 2024-09-25
Revert "#1718 moving Spring boot BOM into submodule deegree-tools-gml and fixing deps"

copierrj:
This reverts commit af1a74d29878642d3655dbaea2642aabedd10699.

Hopefully this change resolves our failing build.


copierrj:
This PR was only created to test a theory that turned out not to be true. Therefore this PR can be closed.

PR #1747, opened 2024-09-25
Revert "Fix config template for MS SQL Server in the deegree web console"

dstenger:
Reverts deegree/deegree3#1734

Hopefully this change resolves our failing build.
If not, this pull request can be closed as invalid.


dstenger:
The reverted commit is not related to the failing build.

PR #1749, opened 2024-10-09
Remove remote layer pointing to external URL from deegree-workspace-utah

dstenger:
Remote layer pointing to https://mrdata.usgs.gov/services/sgmc2?service=WMS&version=1.3.0&request=GetCapabilities was removed from deegree-workspace-utah.

The background is that external dependencies shall be removed leading to possible unexpected build failures (as it was the case last week).


dstenger:
A second remote ows is still configured in the workspace: https://github.com/deegree/deegree3/blob/main/deegree-workspaces/deegree-workspace-utah/src/main/resources/datasources/remoteows/cascade_demo_deegree_wms.xml

stephanr (approved):

tfr42 (approved):
Ok to remove this demo instance

PR #1757, opened 2024-11-06
Fix recognization of deprecated FeatureCollections in deegree

### LINKED ISSUE #1752: Recognization of FeatureCollections in deegree
## Background
deegree GML tools CLI is used to generate configuration from an INSPIRE schema. However, the generated configuration is empty. We analyzed this and the result was that some FeatureTypes are recognized as FeatureCollections. There is no configuration written for FeatureCollections.
The problem occurred with INSPIRE schema Geophysics and FeatureTypes `ge_gp:GeophProfile` and `ge_gp:Campaign`.

## What does the [GML specification](https://portal.ogc.org/files/?artifact_id=20509) say?
The specification says: "A GML feature collection is any GML feature with a property element in its content model whose content model is derived by extension from gml:AbstractFeatureMemberType (see 9.9.2)" 

## How does deegree behave?
FeatureTypes are considered as FeatureCollection in method ​[isGML32FeatureCollection()](https://github.com/deegree/deegree3/blob/f227d2c29c677c72fbf266d8fdf015caec3ed3d8/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/schema/GMLSchemaInfoSet.java#L550) if they have a property which is either derived from "AbstractFeatureMemberType" or "FeaturePropertyType".
There is also following comment in the code:
```
// handle deprecated FeatureCollection types as well (their properties are not
// based on
// AbstractFeatureMemberType, but on FeaturePropertyType)
```

## How is this use case (INSPIRE) handled by deegree?
The INSPIRE schema has properties which are derived from "gml:FeaturePropertyType" and, thus, those FeatureTypes are considered as FeatureCollections by deegree.
In this case, the property "sam:sampledFeature" defined by https://schemas.opengis.net/sampling/2.0/samplingFeature.xsd causes the problem as it is of type "gml:FeaturePropertyType":
```
...
<element name="sampledFeature" type="gml:FeaturePropertyType" maxOccurs="unbounded" nillable="true" block="#all">
...
```

## Why is deegree behaving like this?
The "AbstractFeatureCollectionType" is defined in ​https://schemas.opengis.net/gml/3.2.1/deprecatedTypes.xsd and includes a sequence of elements of type "FeaturePropertyType" (element featureMember):
```
...
<element name="featureMember" type="gml:FeaturePropertyType">
  <annotation>
    <appinfo>deprecated</appinfo>
  </annotation>
</element>
...
<complexType name="AbstractFeatureCollectionType" abstract="true">
  <annotation>
    <appinfo>deprecated</appinfo>
  </annotation>
  <complexContent>
    <extension base="gml:AbstractFeatureType">
      <sequence>
        <element ref="gml:featureMember" minOccurs="0" maxOccurs="unbounded"/>
        <element ref="gml:featureMembers" minOccurs="0"/>
      </sequence>
    </extension>
  </complexContent>
</complexType>
<element name="AbstractFeatureCollection" type="gml:AbstractFeatureCollectionType" abstract="true" substitutionGroup="gml:AbstractFeature">
...
```
So, the deprecatedTypes schema gives a hint why deegree behaves like this.

## What are possible solutions?
 - Update method isGML32FeatureCollection(): Recognization of FeatureCollections basing on FeaturePropertyType for all types marked with 'deprecated'.
 - Update method isGML32FeatureCollection(): Recognization of FeatureCollections basing on FeaturePropertyType only for types in GML namespace.
 - Update method isGML32FeatureCollection(): Recognization of FeatureCollections basing on FeaturePropertyType not for types in INSPIRE namespace.
 - Allow FeatureCollections when writing SQLFeatureStore configuration generally (configurable via parameter).

lgoltz:
This PR fixes #1752 by introducing a new tunable `deegree.gml.parse.recognize-deprecated-types`. If the tunable is set to true (default) the feature types, not in GML 3.2 namespace with at least one property derived from FeaturePropertyType, are recognized as feature collection. If false only feature types in GML 3.2 namespace with at least one property derived from FeaturePropertyType are recognized as feature collection.

### DIFF
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/schema/GMLSchemaInfoSet.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/gml/schema/GMLSchemaInfoSet.java
@@ -45,6 +45,7 @@ import static org.deegree.commons.tom.gml.GMLObjectCategory.FEATURE;
 import static org.deegree.commons.tom.gml.GMLObjectCategory.GEOMETRY;
 import static org.deegree.commons.tom.gml.GMLObjectCategory.TIME_OBJECT;
 import static org.deegree.commons.tom.gml.GMLObjectCategory.TIME_SLICE;
+import static org.deegree.commons.utils.TunableParameter.get;
 import static org.deegree.commons.xml.CommonNamespaces.GML3_2_NS;
 import static org.deegree.commons.xml.CommonNamespaces.GMLNS;
 import static org.deegree.commons.xml.CommonNamespaces.ISOAP10GMDNS;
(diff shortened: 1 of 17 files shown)


tfr42:
PR #1758 for 3.6 line (`main` branch) has been merged.

PR #1765, opened 2024-12-09
Improvements for writing GeoJson for GML with nested geometries

stempler:
These changes are related to issues we encountered specifically with GetFeatureInfo requests for INSPIRE Governmental Services data using GeoJson format. The solution probably also covers other GML application schemas with a similar structure for defining geometries.

For these cases the WMS would yield an internal error without helpful message as result for the GFI request.
In the logs error messages similar to this would show up:

```
java.lang.IllegalStateException: Dangling name: serviceLocationByGeometry
```

The first commit attempts to solve this problem, by not writing the incomplete Json field, which before resulted in the shown error, as no value was written for the field.

The second commit attempts to retrieve the geometry properties in such a schema (previously it would not identify them), so when the GFI is requested including geometries, they can be included instead of having a `null` value for the geometry. This has the same restriction as before, that if there are multiple geometries the GeoJson representation can't handle this.


deegree-core/deegree-core-base/src/main/java/org/deegree/geojson/GeoJsonWriter.java
```diff
@@ -219,7 +281,10 @@ else if (propertyType instanceof FeaturePropertyType) {
 			exportFeaturePropertyType(property);
 		}
 		else if (propertyType instanceof GeometryPropertyType) {
-			// Do nothing as geometry was exported before.
+			// geometry was exported before
+			// need to throw an exception here, because if we do nothing the Json is
+			// invalid (condition should be handled before)
+			throw new IOException("Geometry should not be rendered as part of GeoJson properties.");
```
tfr42, line 287: Can you please explain why do you see the need to throw an exception here or is it intended to prevent misuse of the API here? Does throwing the exception stops the generation of the GFI GeoJSON output?

tfr42:
Thank you for your contribution. Your pull request has been discussed by the TMC 2024-12-18. Please see the following remarks from the TMC members inline.
You are welcome to join the next TMC meeting in 2025.

deegree-core/deegree-core-base/src/main/java/org/deegree/geojson/GeoJsonWriter.java
stempler: This code should not be reached, the exception is thrown to raise a failure if it does happen nonetheless.

If such a case occurs and the behavior here would be as before, doing nothing, no valid Json would be created and the process would fail with the "Dangling name" exception mentioned in the PR description, because there would be a Json field w/o a value.

So the behavior would be similar, just a different exception.

Now we catch that case earlier though to prevent the field from being written at all and thus to prevent the exception.

The intention here seemed clearly to not export the geometry again, so I guess only other alternative would be creating a `null` value for the field or an empty object. Depends on what is preferable - I went for omitting the field in that case.

stephanr (approved):

deegree-core/deegree-core-base/src/main/java/org/deegree/geojson/GeoJsonWriter.java
tfr42: @stempler Thanks for the explanation.  Would it be possible that you provide a workspace with a example data set to verify the behavoir in an integration test?

stempler: I think the added unit tests in `GeoJsonFeatureWriterTest` already cover the issue and fix quite well.
The data used in the unit tests could be used in test workspaces as well I guess. Do you have an example of a similar integration test like what you have in mind?

PR #1801, opened 2025-04-03
Upgraded spotbugs-maven-plugin to 4.9.3.0

lgoltz:
This PR fixes issue #1800 by setting the spotbugs-maven-plugin to version 4.9.3.0 to resolve the build issue during pipeline step "Quality Checks".

PR #1802, opened 2025-04-07
Fixed version for spotbugs maven plugin in Jenkins pipeline

### LINKED ISSUE #1800: Build broken on main
Currently the build is broken on `main` branch.
https://buildserver.deegree.org/view/All/job/deegree-3.6-release-pipeline/11/
https://buildserver.deegree.org/view/All/job/pull-request/job/main/127/

tfr42:
This PR fixes #1800 by setting the fixed version 4.8.6.0 for the spotbugs maven plugin to avoid in incompatibility with doxia when executing quality goal in Jenkins pipeline. PR #1801 set the plugin version for maven profile "site-all-reports" only.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,7 +51,7 @@ pipeline {
             }
             steps {
                 echo 'Quality checking'
-                sh 'mvn -B -C -fae -P oracle com.github.spotbugs:spotbugs-maven-plugin:spotbugs javadoc:javadoc'
+                sh 'mvn -B -C -fae -P oracle com.github.spotbugs:spotbugs-maven-plugin:4.8.6.0:spotbugs javadoc:javadoc'
             }
             post {
                 always {


lgoltz (approved):

julianzz98 (approved):

dstenger:
Pull request #1801 sets spotbugs maven plugin version to `4.9.3.0`.
This pull request, however, sets it to version `4.8.6.0`.
Is this difference intended?

stephanr (approved):
As a workaround, to get unstuck, this hack is acceptable but does not solve the underlaying issue.

> Pull request #1801 sets spotbugs maven plugin version to `4.9.3.0`. This pull request, however, sets it to version `4.8.6.0`. Is this difference intended?

I also think this isn't ideal, to have different versions, especially as `spotbugs-maven-plugin` only gets its version if the profile `integration-tests` is activated.
So I would suggest that it should be checked if the version can be controlled outside the profile scope, so it also could be called without full reference.
Maybe `pluginManagement` like `dependencyManagement` could define the dependency version on the root pom.xml, but this has to be checked in the maven documenation.

tfr42:
@dstenger @stephanr The plugin is used in both maven lifecycles "site" and "build". For sure the plugin version used in both phases should be equal. But the maven site hasn't been build since version 3.3, see https://apidoc.deegree.org/.

PR #1814, opened 2025-05-16
Fixed non-transparant WMS output for remote WMS and GDAL layers

copierrj:
This fix re-enables JPEG output and correct handling of the `transparent` and `bgcolor` WMS parameters for remote WMS and GDAL layers.

For both layer types deegree tries to reuse an existing `BufferedImage` (provided by the WMS client or by GDAL) and this fails when this image has an alpha layer and JPEG output is requested or transparency is not enabled in the request.

### DIFF
--- a/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java
+++ b/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java
@@ -79,7 +79,11 @@ public class ImageRenderContext extends Java2DRenderContext {
 		this.info = info;
 	}
 
-	public static RenderContext createInstance(RenderingInfo info, BufferedImage image, OutputStream outputStream) {
+	public static RenderContext createInstance(RenderingInfo info, BufferedImage existingImage,
+			OutputStream outputStream) {
+		BufferedImage image = ImageUtils.prepareImage(existingImage, info.getFormat(), info.getWidth(),
+				info.getHeight(), info.getTransparent(), info.getBgColor());
+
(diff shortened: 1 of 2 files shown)


stephanr:
@copierrj 
After discussion on this topic in todays TMC meeting, I am unsure if this fix won't have a negative impact as its implementation was done with a focus on the GDAL layer.

https://github.com/deegree/deegree3/blob/c94463e79551f660851cdf7349a5a6aa995cd8df/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java#L112-L123

In Line 112, the new behavior is introduced for all cases where "jpeg" **or** all image formats without transparency.

And we already have this kind of logic already present in ImageUtils at:
https://github.com/deegree/deegree3/blob/01ebf3743dccf7d1f706ea983fb67d04468b46f3/deegree-core/deegree-core-style/src/main/java/org/deegree/style/utils/ImageUtils.java#L61-L93

Which is used by one of the possible two constructors:
https://github.com/deegree/deegree3/blob/01ebf3743dccf7d1f706ea983fb67d04468b46f3/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java#L86-L90

So I am asking if the fix may be better located at the code calling the other constructor with an already created image:
https://github.com/deegree/deegree3/blob/c94463e79551f660851cdf7349a5a6aa995cd8df/deegree-core/deegree-core-rendering-2d/src/main/java/org/deegree/rendering/r2d/context/ImageRenderContext.java#L82-L84

copierrj:
I've reworked the PR as suggested by @stephanr. Unfortunately I didn't fully understand the existing implementation regarding png8 and gif output. The solution as provided in this PR makes the most sense to me.

stephanr (approved):

PR #1827, opened 2025-06-18
Fix nil properties in GeoJSON (3.6)

lgoltz:
Currently properties with xsi:nil="true" and without other attributes/elements results in following GeoJSON:
```
  "attributeName": true,
```
This is irritating as the value is empty.
This PR excludes attributes with  xsi:nil="true" from the GeoJSON.

PR with Backport for 3.5: #1826


dstenger (approved):

tfr42:
During the discussion we found that https://github.com/INSPIRE-MIF/gp-geojson-encodings/blob/main/spec/geojson-encoding-rule.md#nilreason-information page follwing the same rational.

stephanr (approved):

PR #1833, opened 2025-06-29
Fixes GFI on a group layers if one of the sub layers is not queryable (3.6)

### LINKED ISSUE #1822: GFI on a group layers fails if one of the sub layers is not queryable
**Setup**
* RemoteWMS with several RemoteLayers 
* Some of the RemoteLayers are combined to a group of layers 
* One of the child RemoteLayers is not queryable
* deegree 3.5

The GetFeatureInfo request on the group layer fails:

```
<ServiceExceptionReport>
  <ServiceException code="LayerNotQueryable">GetFeatureInfo is requested on a Layer (name: child1) that is not queryable.</ServiceException>
</ServiceExceptionReport>
```

**Expected behaviour:**
The GetFeatureInfo request on a group layer should ignore all sub layers which are not queryable.

lgoltz:
Fixes #1822  Exception is replaced by a log warning.  

Backport to 3.5: #1834

### DIFF
--- a/deegree-services/deegree-services-wms/src/main/java/org/deegree/services/wms/MapService.java
+++ b/deegree-services/deegree-services-wms/src/main/java/org/deegree/services/wms/MapService.java
@@ -334,18 +334,16 @@ public class MapService {
 		ListIterator<LayerQuery> queryIter = queries.listIterator();
 		for (LayerRef n : gfi.getQueryLayers()) {
 			LayerQuery query = queryIter.next();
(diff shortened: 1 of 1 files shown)


julianzz98:
> It should be tested if an exception is thrown when a non-queryable layer (layer attribute queryable="0") is requested via GFI directly. The standard mandates that an exception is returned in this case.

In this case, no exception is being thrown. This is the same behavior as before the bugfix!

PR #1834, opened 2025-06-29
Fixes GFI on a group layers if one of the sub layers is not queryable (3.5)

lgoltz:
Backport for 3.5 of #1833


copierrj:
The TMC discussed this PR and is reluctant to merge this change into a maintenance branch. This behavioral change might break existing installations. Adding a tunable to make this change an opt-in would be appreciated.

stephanr:
Thank you for your contribution. Unfortunately, after carefully reviewing your contribution we, the TMC members, came to the conclusion not to merge it into the deegree code base. Of course this is not the final word. If you think, that we did wrong then please re-initiate the decision making process by making a proposal and start a discussion on the mailing list. Perhaps we can reach a consensus then. Anyhow, we do appreciate your commitment to open source and to deegree in special.

PR #1849, opened 2025-08-06
Fixes nil properties in GeoJSON (3.6)

lgoltz:
This PRs fixes incomplete nil Properties. Example from the test:

Before (the two first new assertions failed):
```
"name" : {
        "GeographicalName" : {
          "sourceOfName" : true,
            "spelling" : {
             ....
            "pronunciation" : true,
            "nameStatus" : {
              "nil" : true
            },
      ...
```

After: 
```
"name" : {
        "GeographicalName" : {
          "sourceOfName" : {
            "nil" : true
          },
          ....
          "pronunciation" : {
            "nil" : true
          },
          ....
          "nameStatus" : {
            "nil" : true
          },
      ...
```

### DIFF
--- a/deegree-core/deegree-core-base/src/main/java/org/deegree/geojson/GeoJsonWriter.java
+++ b/deegree-core/deegree-core-base/src/main/java/org/deegree/geojson/GeoJsonWriter.java
@@ -420,7 +420,15 @@ public class GeoJsonWriter extends JsonWriter implements GeoJsonFeatureWriter, G
 			QName key = attributesAndChildren.keySet().iterator().next();
 			List<TypedObjectNode> values = attributesAndChildren.get(key);
 			if (values.size() == 1 && values.get(0) instanceof PrimitiveValue) {
-				exportValue((PrimitiveValue) values.get(0));
+				if (isNilled(key, values.get(0))) {
+					beginObject();
+					name(key.getLocalPart());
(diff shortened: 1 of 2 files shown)


stephanr:
After discussing this PR in today's TMC meeting we stumbled upon the different implementation in comparison to https://github.com/deegree/deegree3/pull/1827 .
There https://github.com/INSPIRE-MIF/gp-geojson-encodings/blob/main/spec/geojson-encoding-rule.md#nilreason-information was mentioned that stated to omit these properties.

Here I assume the used dataset was https://inspire.ec.europa.eu/schemas/gn/4.0/GeographicalNames.xsd (or another version) in which the property `sourceOfName` is of type string.

So we are replacing the export of nil from a boolean of `true` to an object of `{"nil": true}` which both does not really match the intended datatype of `string` (in the XSD schema).

Technically, there are multiple choices:
- `true` (obviously not a good idea)
- `false` (obviously not a good idea)
- `{}` object with or without data
- `[]` array with or without data
- `undefined` see Chapter 5 JSON Value of ECMA-404_2nd_edition_december_2017
- omit the property completely, similar to https://github.com/INSPIRE-MIF/gp-geojson-encodings/blob/main/spec/geojson-encoding-rule.md#nilreason-information 

For today we postponed this topic and would appreciate any input.

stephanr (approved):

PR #1947, opened 2026-04-27
Updated dependencies for JAXB, Jaxen, GDAL, Faces, Primefaces and Spring Boot

tfr42:
This PR upgrades the dependencies:
- jaxb-runtime
- jakarta.faces
- jaxen
- gdal
- primefaces and  primefaces-extensions
- Spring Boot to 3.5.14 (which is the latest and last release of the 3.x line, see https://spring.io/projects/spring-boot#support)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -504,7 +504,7 @@
       <dependency>
         <groupId>org.glassfish.jaxb</groupId>
         <artifactId>jaxb-runtime</artifactId>
-        <version>4.0.6</version>
+        <version>4.0.7</version>
         <scope>runtime</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1205,7 +1205,7 @@
       <dependency>
         <groupId>org.gdal</groupId>
         <artifactId>gdal</artifactId>
-        <version>3.11.0</version>
+        <version>3.12.0</version>
```
dstenger, line 1208: Handbook still mentions version 3.11.0: https://download.deegree.org/documentation/3.6.7/html/#_connecting_gdal_and_deegree

This should be updated as well.

In general, we can think about putting this dependency to scope provided. Then users must add this dependency to classpath by themselves.

julianzz98:
The deegree-tools-gml were successfully tested under this version.

stephanr (approved):

PR #1961, opened 2026-06-17
Fixed NumberFormatException if sqlValue is to long for java integer range

lgoltz:
Currently a request with a 17-digit value results in a NumberFormatException. This PR  fixed this by parsing the value as long as fallback.
Besides the unused logger was removed.


copierrj:
Looks good to me. Perhaps we could (some day) investigate if using something like [NumberUtils](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html) would be useful.

copierrj (approved):

stephanr (approved):

dstenger (approved):

tfr42 (approved):

PR #1970, opened 2026-07-23
Bumped postgres and sqlserver driver (3.5)

tfr42:
Bumped to latest version:
- org.postgresql:postgresql:42.7.13
- com.microsoft.sqlserver:mssql-jdbc:12.10.2.jre11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -366,6 +366,7 @@
                     <ignoreClass>com.sun.media.jai.*</ignoreClass>
                     <ignoreClass>com.sun.xml.*</ignoreClass>
                     <ignoreClass>org.w3c.dom.*</ignoreClass>
+                    <ignoreClass>org.deegree.commons.metadata.description.jaxb.*</ignoreClass>
                   </ignoreClasses>
                   <scopes>
                     <scope>compile</scope>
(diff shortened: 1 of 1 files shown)


julianzz98 (approved):

pom.xml
```diff
@@ -366,6 +366,7 @@
                     <ignoreClass>com.sun.media.jai.*</ignoreClass>
                     <ignoreClass>com.sun.xml.*</ignoreClass>
                     <ignoreClass>org.w3c.dom.*</ignoreClass>
+                    <ignoreClass>org.deegree.commons.metadata.description.jaxb.*</ignoreClass>
```
dstenger, line 369: Why is this ignore necessary?

dstenger: One of the dependencies probably introduces a duplicate jaxb classes. Is it safe that this will not lead to problems?

tfr42: On local build this was blocking, no time to investigate what the root cause is. Build requires JDK 11 with Maven 3.8. The class generated during build is neither part of postgresql nor mssql-jdbc nor do the jdbc driver contain class in the org.deegree packages.

PR #1975, opened 2026-08-06
Enhanced SldProvider to allow specific implementations retrieving values for SLD parameter

lgoltz:
This PR enhances the parsing of the sld parameter. Currently only parsing from a  http url is supported. This PR adds a SldProvider which can be implemented to allow specific access to SLD files, e.g. from S3 .


dstenger (approved):

copierrj:
TMC discussed this topic in length but didn't reach a final conclusion yet. One thing that came up was the question of how to facilitate both novices and expert users.

We are leaning to not using ServiceLoader for this functionality but reaching for a tunable param instead. Such a param could provide a implementation class to be used by the WMS.

By providing multiple sensible 'stock' implementations (e.g. DenyAll, AllowAll) for the user to choose from and provide an extension point for system integrator to plug-in their own functionality we serve both kinds of users.

The TMC planned to revisit this topic in the next TMC meeting. Others are kindly invited to comment on this issue to share their point of view.

